PR #35, opened 2020-12-02
fix dataset update endpoint according to Native API spec

joenio:
reference:
https://guides.dataverse.org/en/latest/api/native-api.html#publish-a-dataset


tainguyenbui (approved):

tainguyenbui:
thanks @joenio for fixing this endpoint 🚀

PR #67, opened 2023-06-13
65 - Add CitationMetadataBlock type description

### LINKED ISSUE #65: Modify the dataset type description to explicitly declare the citation metadata fields
Modify the dataset type definition to explicitly declare the citation metadata fields

## What inspired the request?
- https://github.com/IQSS/dataverse-frontend/issues/115

We'll be using some fields of the citation metadata for different behaviours of the SPA, it makes sense that the CitationMetadata fields are explicitly defined in the dataset type so the frontend developers can know what to expect from the citation. For example the dataset title it's a field that no matter the frontend, it will be handled with some specific UI logic, so it doesn't make sense to have this title hidden behind a generic MetadataField type without knowing the key of this field. 

There are more cases like this one so we decided to make the citation block fields explicit in the definition of the dataset interface so any user of the js-dataverse module knows what keys to expect from the dataset citation metadata block.

Add the specific keys of the citation metadata fields to the Dataset interface definition

- https://github.com/IQSS/dataverse-frontend/issues/115

MellyGray:
This PR adds the CitationMetadataBlock interface to define the expected citation metadata fields. This way any user accessing the dataset data through the js-dataverse module knows what to expect from the citation metadata fields.

We need this because the Citation Metadata contains required metadata fields, so they need to be declared in order to create a new dataset using the js-dataverse module. And in order to declare those fields, the user needs to know which are these required fields.

This does not apply to the other metadata blocks because the other metadata blocks are optional

- Closes #65 

This PR doesn't add any modification to the behaviour of the use cases, it just adds a type definition, so everything should be working as it used to

1. npm run build
2. Go to dist/datasets/domain/models/Dataset.d.ts
3. You can see that the type definition includes the the definition of the Citation Metadata fields so any user of the js-dataverse module is now able to know what to expect from this metadata block. `export type DatasetMetadataBlocks = [CitationMetadataBlock, ...DatasetMetadataBlock[]];`

No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@types/node": "^18.15.11",
-        "@types/turndown": "5.0.1",
+        "@types/turndown": "^5.0.1",
         "axios": "^1.3.4",
         "turndown": "^7.1.2",
         "typescript": "^4.9.5"
--- a/src/core/domain/useCases/UseCase.ts
+++ b/src/core/domain/useCases/UseCase.ts
@@ -1,3 +1,4 @@
 export interface UseCase<T> {
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
   execute(...args: any[]): Promise<T>;
 }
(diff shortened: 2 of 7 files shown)


ekraffmiller (approved):
I ran the local build and I can see the CitationMetadataBlock in Dataset.d.ts. Approved 👍

PR #69, opened 2023-06-21
63 - Add use case for getting metadata blocks

### LINKED ISSUE #63: Add use case to get the metadata blocks info
Add a new use case to the module to retrieve the metadata blocks info, including the dataset field types info. 

## What inspired the request?
- https://github.com/IQSS/dataverse-frontend/issues/115

Since the metadata blocks are described by the tsvs, including the fieldType and the displayFormat, they should be retrieved by the js-dataverse module so the frontend can read the tsvs definitions and display the metadata fields in the proper way, this includes for example which separation the compound fields should use ';', '()', etc

This may need to extend the API since I tested this [endpoint](https://guides.dataverse.org/en/latest/api/native-api.html#metadata-blocks-api) but it does not retrieve the information about the displayFormat

## What existing behavior do you want changed?
None

Add a use case to get the tsvs of the system, for example `getMetadataBlockInfo('citation').execute()`.

- https://github.com/IQSS/dataverse-frontend/issues/115

GPortas:

Adds a new use case to retrieve the metadata blocks info, including the dataset field types info.

- Closes #63 

New changes require https://github.com/IQSS/dataverse/pull/9668 to work

Changes are covered with tests, although the new use case should be tested from the SPA and refined if necessary.

Integration tests are pending (as for the other use cases of the module). They will be included once the integration test environment is established.

New use case to retrieve the metadata blocks info

## Additional documentation:

N/A


pdurbin (approved):
I didn't test this but it makes sense. Approved.

PR #80, opened 2023-07-20
Add use cases for file tab data visualization

### LINKED ISSUE #77: [Spike] [1/2] Add use case logic to get display file data for the frontend files tab
Add use case logic to support the frontend files tab display data.

The new logic will need the API changes in https://github.com/IQSS/dataverse/issues/9692 to work.

## What inspired the request?
- https://github.com/IQSS/dataverse-frontend/issues/104
- https://github.com/IQSS/dataverse/issues/9692 

## What existing behavior do you want changed?
None

Add new use cases for getting files data

- https://github.com/IQSS/dataverse-frontend/issues/104
- https://github.com/IQSS/dataverse/issues/9692

GPortas:

Adds the use case logic to support the frontend files tab display data.

The new use case logic needs the API changes implemented in https://github.com/IQSS/dataverse/issues/9692 to work.

In particular, this PR adds the following new use cases to the package:

- getDatasetFiles
- getFileDownloadCount
- getFileDataTables
- getUserPermissionsOnFile

In addition, the dataset use cases have been refactored. In particular, the old getDatasetById and getDatasetByPersistentId use cases are merged into a single getDataset use case, which is inferring the type of id specified.

- Closes #77

New use cases for files:

- getDatasetFiles
- getFileDownloadCount
- getFileDataTables
- getUserPermissionsOnFile


src/files/domain/models/File.ts
```diff
@@ -0,0 +1,39 @@
+export interface File {
```
MellyGray, line 1: Why the File interface does not include the user permissions? If I understand the code, the only way to get the permissions is by calling getFileUserPermissionsByFileId? Is there a way to get all the files permissions at once?

Permissions information is key in the frontend since it's used on many conditional renders and info messages.

I don't want to couple the js-dataverse implementation to the frontend but I'm just warning you about what is going to happen in the frontend if the permissions can only be retrieved 1 by 1. When the user opens the Dataset View mode there's going to be 1 request to getDatasetFiles and from 10 requests to getFileUserPermissions up to 50 requests if the pageSize is 50.

I mean, when the getDatasetFiles use case is called the API already knows who is the user requesting the files so why not include the permissions in the response? If not, maybe we can include a use case to getAllFileUserPermissionsByDatasetId

I'm proposing some solutions because I'm concerned about overloading the getFileUserPermissions endpoint, but maybe you already had some idea on how to manage this in the frontend

GPortas: Thanks for your message. We've already taken into account the endpoint overload concerns you mention, I suggest you to take a look at this thread: https://github.com/IQSS/dataverse/pull/9693#discussion_r1284130853

 Let's catchup, it is going to be more agile.

src/core/infra/repositories/ApiRepository.ts
```diff
@@ -4,9 +4,11 @@ import { ReadError } from '../../domain/repositories/ReadError';
 import { WriteError } from '../../domain/repositories/WriteError';
 
 export abstract class ApiRepository {
-  public async doGet(apiEndpoint: string, authRequired = false): Promise<AxiosResponse> {
+  DATASET_VERSION_LATEST = ':latest';
```
MellyGray, line 7: ApiRepository shouldn't be couple to a specific endpoint right? I mean, why does ApiRepository know about DATASET_VERSION_LATEST? 

I would move this constant somewhere else closer to the dataset endpoints to avoid coupling ApiRepository to specific endpoints

src/files/infra/repositories/FilesRepository.ts
```diff
@@ -0,0 +1,92 @@
...
+  public async getDatasetFiles(
+    datasetId: number | string,
+    datasetVersionId?: string,
+    limit?: number,
+    offset?: number,
+    orderCriteria?: FileOrderCriteria,
+  ): Promise<File[]> {
+    if (datasetVersionId === undefined) {
+      datasetVersionId = this.DATASET_VERSION_LATEST;
+    }
```
MellyGray, line 27: I think you can avoid this if by giving a default value to datasetVersionId arg
```suggestion
  public async getDatasetFiles(
    datasetId: number | string,
    datasetVersionId: string = this.DATASET_VERSION_LATEST,
    limit?: number,
    offset?: number,
    orderCriteria?: FileOrderCriteria,
  ): Promise<File[]> {
```

src/files/infra/repositories/FilesRepository.ts
```diff
@@ -0,0 +1,92 @@
...
+    const queryParams: GetFilesQueryParams = {};
+    if (limit !== undefined) {
+      queryParams.limit = limit;
+    }
+    if (offset !== undefined) {
+      queryParams.offset = offset;
+    }
+    if (orderCriteria !== undefined) {
+      queryParams.orderCriteria = orderCriteria.toString();
+    }
```
MellyGray, line 43: You can apply this same syntax here for the queryParams, right?

{
...(fileDataTablePayload.varQuantity && { varQuantity: fileDataTablePayload.varQuantity }),
...(fileDataTablePayload.caseQuantity && { caseQuantity: fileDataTablePayload.caseQuantity }),
...(fileDataTablePayload.recordsPerCase && { recordsPerCase: fileDataTablePayload.recordsPerCase }),
UNF: fileDataTablePayload.UNF,
dataVariables: transformPayloadToDataVariables(fileDataTablePayload.dataVariables),
};

test/integration/datasets/DatasetsRepository.test.ts
```diff
@@ -35,57 +26,58 @@ describe('DatasetsRepository', () => {
...
-
-    test('should return dataset when it exists filtering by id and version id', async () => {
-      const actual = await sut.getDatasetById(createdTestDatasetId, ':draft');
-      expect(actual.id).toBe(createdTestDatasetId);
-    });
+  describe('getDataset', () => {
+    describe('by numeric id', () => {
+      test('should return dataset when it exists filtering by id', async () => {
+        const actual = await sut.getDataset(TestConstants.TEST_CREATED_DATASET_ID);
+        expect(actual.id).toBe(TestConstants.TEST_CREATED_DATASET_ID);
```
MellyGray, line 33: Why do you check only the id instead of the whole Dataset? I mean, checking all the properties from Dataset.ts

test/integration/datasets/DatasetsRepository.test.ts
```diff
@@ -35,57 +26,58 @@ describe('DatasetsRepository', () => {
...
+      });
 
-    test('should return error when dataset does not exist', async () => {
-      let error: ReadError = undefined;
-      await sut.getDatasetById(nonExistentTestDatasetId).catch((e) => (error = e));
+      test('should return dataset when it exists filtering by id and version id', async () => {
+        const actual = await sut.getDataset(TestConstants.TEST_CREATED_DATASET_ID, ':draft');
+        expect(actual.id).toBe(TestConstants.TEST_CREATED_DATASET_ID);
+      });
```
MellyGray, line 39: This makes me think that we could change the use case definition to `getDataset(datasetId: number | string, datasetVersionId?: string | DatasetNotNumberedVersion): Promise<Dataset>`

enum DatasetNotNumberedVersion {
draft: ":draft",
latest: ":latest"
}

just so the user of the module knows that datasetVersionId can be latest or draft

test/integration/files/FilesRepository.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        assert.match(actual[1].name, testTextFile2Name);
+        assert.match(actual[2].name, testTextFile3Name);
+        assert.match(actual[3].name, testTabFile4Name);
+      });
+
+      test('should return correct files filtering by dataset id and paginating', async () => {
+        const actual = await sut.getDatasetFiles(TestConstants.TEST_CREATED_DATASET_ID, undefined, 3, 3, undefined);
+        assert.match(actual.length, 1);
+        assert.match(actual[0].name, testTabFile4Name);
+      });
```
MellyGray, line 73: I'd add 2 different tests for the limit and the offset as well

sut.getDatasetFiles(TestConstants.TEST_CREATED_DATASET_ID, undefined, undefined, 3, undefined)

sut.getDatasetFiles(TestConstants.TEST_CREATED_DATASET_ID, undefined, 3, undefined, undefined)

Also, I was thinking if they should be an object since they're so tied together that there's a test to test them together

paginating {
 limit?: number,
     offset?: number
}

test/unit/files/FilesRepository.test.ts
```diff
@@ -0,0 +1,479 @@
...
+
+        actual = await sut.getFileDownloadCount(testFile.id);
+
+        assert.calledWithExactly(
+          axiosGetStub,
+          expectedApiEndpoint,
+          TestConstants.TEST_EXPECTED_AUTHENTICATED_REQUEST_CONFIG_SESSION_COOKIE,
+        );
+        assert.match(actual, testCount);
+      });
```
MellyGray, line 228: seems like a repeating pattern

// API Key auth
test

// Session cookie auth
test

I think is worth developing a way of automatically running any test twice with the 2 configuration, maybe in another issue

jest.config.js
```diff
@@ -7,4 +7,5 @@ module.exports = {
   moduleFileExtensions: ['ts', 'js', 'json', 'node'],
   coveragePathIgnorePatterns: ['node_modules', 'testHelpers'],
   globalSetup: '<rootDir>/test/integration/environment/setup.js',
+  testTimeout: 15000,
```
MellyGray, line 10: are these seconds? Seems like a lot

unit tests time in red, should we check that?

image

MellyGray (commented):
✅ Tests passing 
✅ Code looks good overall (Just a few comments, not critical things so I don't want to block this PR)

src/files/infra/repositories/FilesRepository.ts
GPortas: Yes, but it's significantly less readable and I preferred to use a cleaner approach since there aren't too many parameters, as in the example you shared with me from a "transformer" function.

test/integration/datasets/DatasetsRepository.test.ts
GPortas: Because the response-to-model transformation is already tested in the unit test with all properties checked using mocked responses. I found no need to check the properties also here. In any case any error in the response-to-model transformation would raise an error both in the integration and unit test.

jest.config.js
GPortas: Integration tests can take some time, as multiple Dataverse API calls are made in the integration test setup methods to configure certain scenarios and preconditions in the Dataverse test container instance. 

Initially, I established this timeout by test, where I found it necessary. As I repeated it a lot, I decided to set it globally. Note that this timeout is only set when running the integration test suite (npm test and npm test:integration). When we want to run only the unit tests (by running npm test:unit), the timeout is the default one.

I'm not worried about the timing of the unit tests at the moment (There are many tests per file). I'll keep that in mind more from now, but I think it is not an issue or blocker for this PR.

test/integration/files/FilesRepository.test.ts
GPortas: In a later PR I include the FileCriteria object that groups all the different criteria for getting files. See: https://github.com/IQSS/dataverse-client-javascript/blob/3892ce2018803549910fbec29d4f7a6e6d3af161/src/files/domain/models/FileCriteria.ts

I decided not to test all possible filtering scenarios, combining the optional parameters in different ways as you propose, since all those tests are already implemented on the Dataverse API side. If you are interested, see: https://github.com/IQSS/dataverse/blob/35e454767288d191d496b04a2ee0f6471bebac53/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java#L3300

For js-dataverse, I find enough to check that the parameters sent by the module are correctly interpreted by the API, whatever the combination of them.

test/integration/datasets/DatasetsRepository.test.ts
MellyGray: I see, thanks! 

Well, it would raise an error only in the integration, not in the unit test, because last time there was a property missing in the dataset payload but the unit test wasn't failing because the unit test payload is mocked to be a perfect payload. But the integration test would fail, so that's enough 👍

src/files/infra/repositories/FilesRepository.ts
MellyGray: ok, I was just suggesting it to keep the syntax consistent

test/integration/datasets/DatasetsRepository.test.ts
GPortas: I like this approach.

It is also related to what you say in other comments about the use of "latest" in repositories. It is not the responsibility of the data layer to set ":latest" as the default value, but domain layer, in my opinion.

So we end up with this:

`getDataset(datasetId: number | string, datasetVersionId: string | DatasetNotNumberedVersion = DatasetNotNumberedVersion.LATEST): Promise<Dataset>`

I'm refactoring the code, but I'll include the refactor in the next PR so I don't overextend this one.

Thanks!

test/integration/files/FilesRepository.test.ts
MellyGray: > I decided not to test all possible filtering scenarios, combining the optional parameters in different ways as you propose, since all those tests are already implemented on the Dataverse API side. If you are interested, see: https://github.com/IQSS/dataverse/blob/35e454767288d191d496b04a2ee0f6471bebac53/src/test/java/edu/harvard/iq/dataverse/api/DatasetsIT.java#L3300

Sorry, I was thinking this was a unit test. 

Anyway, in the unit tests we don't have these tests either. If we offer the user the possibility to pass undefined to one of those arguments then I think we should check those cases work, and I'm not thinking on testing the API, I'm just thinking on testing the request creation that could be malformed when passing an undefined to one of the arguments of the use case

src/core/infra/repositories/ApiRepository.ts
MellyGray: ok, I think Guillermo is refactoring this in another PR so I'm closing the thread

src/files/infra/repositories/FilesRepository.ts
MellyGray: @GPortas you are refactoring this in another PR right? I'm closing the thread

MellyGray (approved):
All threads solved! There is some refactoring going on in another PR to not block this one. Moving this to QA!

PR #92, opened 2023-09-27
Add GetDatasetFilesTotalDownloadSize use case

### LINKED ISSUE #86: Add use case to get the total space consumed by the files of a particular dataset version

We need a use case to get the total space consumed by the files of a particular dataset version.

This use case will allow the SPA to know the total space consumed by the files when the user selects "select all" in the files tab, without needing to retrieve all the files (without pagination, could be an expensive operation) and apply client side logic to calculate the size.

The use case may be useful for other scenarios where the total file size may be useful, for example, metrics-related applications.

We need to explore the Dataverse API to determine if an endpoint exists to support the new use case or if it needs to be created (Create API spike issue in such case).

Package consumer

- Discussion on https://github.com/IQSS/dataverse-frontend/pull/156

None

New use case for getting the total space consumed by the files of a particular dataset version

- https://github.com/IQSS/dataverse-frontend/pull/156

GPortas:

Adds new GetDatasetFilesTotalDownloadSize use case to support SPA functionality: https://github.com/IQSS/dataverse-frontend/issues/170

- Closes #86 

## Related Dataverse PRs:

- Depends on https://github.com/IQSS/dataverse/pull/9960

Add GetDatasetFilesTotalDownloadSize use case

## Additional documentation:

None


src/files/domain/models/FileDownloadSizeMode.ts
```diff
@@ -0,0 +1,5 @@
+export enum FileDownloadSizeMode {
+  ALL = 'All',
+  ORIGINAL = 'Original',
+  ARCHIVAL = 'Archival',
```
MellyGray, line 4: Just a quick question, so archival is with the tabular files processed, original is with the tabular files not processed and ALL I assume it means both ORIGINAL and ARCHIVAL tabular files, is it?

MellyGray (dismissed):
LGTM!

src/files/domain/models/FileDownloadSizeMode.ts
GPortas: That's it, as you describe!

kcondon:
@GPortas Merge conflict, otherwise github actions test seems to pass.

PR #97, opened 2023-10-12
Extend GetDatasetFilesTotalDownloadSize use case to support FileSearchCriteria and deaccessioned datasets

### LINKED ISSUE #94: Extend GetDatasetFilesTotalDownloadSize use case to support FileSearchCriteria

We need to extend the existing GetDatasetFilesTotalDownloadSize use case to support filters and search criteria.

These changes depend on https://github.com/IQSS/dataverse/issues/9995

Package consumer

- SPA needs

None

Extend the GetDatasetFilesTotalDownloadSize use case to support filters and search criteria

- https://github.com/IQSS/dataverse/issues/9995

GPortas:

Extends current GetDatasetFilesTotalDownloadSize use case to support deaccessioned datasets and search filter criteria.

This logic is necessary to support the SPA files tab.

- Closes #94 

## Related Dataverse PRs:

- Depends on https://github.com/IQSS/dataverse/pull/10014

N/A

## Additional documentation:

None


src/files/domain/useCases/GetDatasetFilesTotalDownloadSize.ts
```diff
@@ -14,11 +15,15 @@ export class GetDatasetFilesTotalDownloadSize implements UseCase<number> {
...
     fileDownloadSizeMode: FileDownloadSizeMode = FileDownloadSizeMode.ALL,
+    fileSearchCriteria?: FileSearchCriteria,
+    includeDeaccessioned: boolean = false,
   ): Promise<number> {
     return await this.filesRepository.getDatasetFilesTotalDownloadSize(
       datasetId,
       datasetVersionId,
+      includeDeaccessioned,
       fileDownloadSizeMode,
+      fileSearchCriteria,
```
MellyGray, line 26: Why is the order of the arguments different in the use case than in the filesRepository? 

it's not that important but if you choose one order why not staying with it?

src/files/infra/repositories/FilesRepository.ts
```diff
@@ -23,6 +23,16 @@ export interface GetFilesQueryParams {
...
+export interface GetFilesTotalDownloadSizeQueryParams {
+  includeDeaccessioned: boolean;
+  mode?: string;
+  contentType?: string;
+  accessStatus?: string;
+  categoryName?: string;
+  tabularTagName?: string;
+  searchText?: string;
+}
+
```
MellyGray: You could have a common interface with the search criteria properties and then extend it with the other properties to not repeat the same properties in both interfaces when we already know those properties are the search criteria

```
export interface GetFilesQueryParams extends SearchCriteriaQueryParams {
  includeDeaccessioned: boolean;
  limit?: number;
  offset?: number;
  orderCriteria?: string;
}

export interface GetFilesTotalDownloadSizeQueryParams extends SearchCriteriaQueryParams {
  includeDeaccessioned: boolean;
  mode?: string;
}
```

MellyGray (commented):
Looks good! 

I just left a few suggestions, take a look at them and I think we can move this forward quickly

src/files/domain/useCases/GetDatasetFilesTotalDownloadSize.ts
GPortas: The order is different because in the repository method the _includeDeaccessioned_ parameter no longer has a default value, since this is determined by the use case.

I cannot put _includeDeaccessioned_ as the last parameter in the repository method since TypeScript does not allow putting a non-optional parameter followed by an optional one (_fileSearchCriteria_ is optional)

I want to leave _includeDeaccessioned_ as the last parameter in the use case so that we don't have to specify it when calling the use case if not necessary.

src/files/infra/repositories/FilesRepository.ts
GPortas: Done!

MellyGray (approved):
Thanks for addressing the changes! LGTM

PR #103, opened 2023-11-15
Optimize GetDataset use case by using includeFiles=false

### LINKED ISSUE #100: Optimize getDataset use case to use the new API query param includeFiles=false to avoid unnecessary file lookups

A new optimization has been recently added to the getVersion endpoint, which is the API endpoint used by the getDataset use case: https://github.com/IQSS/dataverse/blob/develop/src/main/java/edu/harvard/iq/dataverse/api/Datasets.java#L489.

A new query parameter "includeFiles" has been included that supports the option of excluding the file metadatas lookup from the endpoint search. Since we are only interested in the dataset info and not the files metadata (which we recover from other endpoint) we want to use the new query param set to false.

Package consumer

- Recent API optimization

None

None

None

GPortas:

Adds includeFiles=false query parameter to the repository method used by the GetDataset use case, since files are ignored and we don't need their lookup in the backend.

- Closes #100 

## Related Dataverse PRs:

None

### DIFF
--- a/src/datasets/infra/repositories/DatasetsRepository.ts
+++ b/src/datasets/infra/repositories/DatasetsRepository.ts
@@ -36,6 +36,7 @@ export class DatasetsRepository extends ApiRepository implements IDatasetsReposi
       true,
       {
         includeDeaccessioned: includeDeaccessioned,
+        includeFiles: false
       },
     )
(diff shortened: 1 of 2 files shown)


MellyGray (approved):
Looks good!

PR #112, opened 2024-01-03
Modify getDataset repository API call to use 'excludeFiles' parameter

### LINKED ISSUE #111: Modify Dataverse API calls to use 'excludeFiles' parameter
In this Dataverse PR the Dataverse API has changed to use 'excludeFiles' instead of 'includeFiles'.  So the library calls to Dataverse need to use this new parameter.
https://github.com/IQSS/dataverse/pull/10191

For the SPA, we need to be able to show the View Dataset page for the Deaccessioned dataset, whether or not the user has edit permissions on the dataset.  So we can get the dataset by using parameters includeDeaccessioned=true & excludeFiles=true.

From the Dataverse ticket: 

> We agreed to change the parameter name from includeFiles with excludeFiles, this is the default and the lack of the parameter will default to false so files are expected to be included by default if this parameter is not provided.
> 
> getVersionFiles, getVersionFileCounts and getDownloadSize act exactly as before, this check would execute every time includeDeaccessioned was set to true, getDatasetVersionCitation will skip this test when includeDeaccessioned is true and getVersion will check for when includeDeaccessioned is true and excludeFiles is false. This change also replaced includeFiles with excludeFiles which is the default behaviour.
> 
> This change required to update Datasets.getDatasetVersionOrDie which impacted the following methods:
> 
> getVersion {id}/versions/{versionId: Will check file permissions when files when excludeFiles and includeDeaccessioned are true.
> getVersionFiles {id}/versions/{versionId}/files: Will check file permissions when includeDeaccessioned is true.
> getVersionFileCounts {id}/versions/{versionId}/files/counts: Will check file permissions when includeDeaccessioned is true.
> getDownloadSize {identifier}/versions/{versionId}/downloadsize: Will check file permissions when includeDeaccessioned is true.
> getDatasetVersionCitation {id}/versions/{versionId}/citation: Won't check file permissions.
> listVersions {id}/versions: replaced includeFiles with excludeFiles which is the default.

GPortas:

Modifies getDataset repository API call to use 'excludeFiles' parameter.

Also adds more test cases for deaccessioned datasets and getDataset repository method.

- Closes #111

## Related Dataverse PRs:

- None

- Run integration tests or review GitHub action execution

### DIFF
--- a/test/integration/datasets/DatasetsRepository.test.ts
+++ b/test/integration/datasets/DatasetsRepository.test.ts
@@ -25,32 +25,40 @@ describe('DatasetsRepository', () => {
 
   describe('getAllDatasetPreviews', () => {
     const testPageLimit = 1;
+    const expectedTotalDatasetCount = 3;
 
     test('should return all dataset previews when no pagination params are defined', async () => {
       const actual: DatasetPreviewSubset = await sut.getAllDatasetPreviews();
-      assert.match(actual.datasetPreviews.length, 2);
-      assert.match(actual.datasetPreviews[0].title, 'Second Dataset');
-      assert.match(actual.totalDatasetCount, 2);
+      assert.match(actual.datasetPreviews.length, expectedTotalDatasetCount);
+      assert.match(actual.datasetPreviews[0].title, 'Third Dataset');
+      assert.match(actual.totalDatasetCount, expectedTotalDatasetCount);
     });
 
     test('should return first dataset preview page', async () => {
       const actual = await sut.getAllDatasetPreviews(testPageLimit, 0);
       assert.match(actual.datasetPreviews.length, 1);
-      assert.match(actual.datasetPreviews[0].title, 'Second Dataset');
-      assert.match(actual.totalDatasetCount, 2);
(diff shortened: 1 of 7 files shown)


test/integration/datasets/DatasetsRepository.test.ts
ekraffmiller, line 1: Please add a test to get the deaccessioned dataset when the user has logged out. That is the use case that we need to support in the SPA.

ekraffmiller (changes requested):
Hi Guillermo, I tried this PR against the SPA, and it is still not getting the deaccessioned dataset when the user has logged out. Can you add a test for that?

GPortas:
@ekraffmiller 

I just added a test for the scenario you mention, but it is failing due to the API issue I explain on `dv-tech`.

GPortas:
Waiting for: https://github.com/IQSS/dataverse/issues/10207

ekraffmiller (approved):
looks good!

MellyGray (approved):
QA ✅

PR #114, opened 2024-01-10
Adds create dataset use case

### LINKED ISSUE #110: Create use case to create a dataset

The goal is to create a use case for creating a dataset.

This use case will be consumed from Dataverse SPA, although it should be applicable to other scenarios.

The use case should handle client-side data validation

- Use case consumer

- https://github.com/IQSS/dataverse-frontend/issues/231

None

Create Dataset use case

- https://github.com/IQSS/dataverse-frontend/issues/231

GPortas:

This PR includes the createDataset use case. This new use case allows to create a dataset in Dataverse, given a NewDataset model and the collection id where the new dataset will be created.

Before calling the repository creation operation, the use case performs validation of the metadata fields of the different metadata blocks of the new dataset.

This validation throws validation errors of different subtypes, depending on the nature of the error. The most notable aspect of this error handling design is that from the errors generated, we can know which particular field is invalid. Given this information, this mechanism will allow error messages to be dynamically rendered within a form, such as the dataset creation form in the SPA.

- Closes #110

## Related Dataverse PRs:

- Depends on https://github.com/IQSS/dataverse/pull/10225

This is the first resource creation operation in the module after the restructuring.

Check integration tests on GitHub actions or in local runs.


test/integration/environment/.env
```diff
@@ -1,6 +1,6 @@
 POSTGRES_VERSION=13
 DATAVERSE_DB_USER=dataverse
 SOLR_VERSION=9.3.0
-DATAVERSE_IMAGE_REGISTRY=docker.io
-DATAVERSE_IMAGE_TAG=unstable
+DATAVERSE_IMAGE_REGISTRY=ghcr.io
+DATAVERSE_IMAGE_TAG=10216-metadatablocks-payload-extension
```
GPortas: This will be reverted once https://github.com/IQSS/dataverse/pull/10225 is merged into dataverse develop. For now we can test the PR with this PR-generated image. Before merging this PR it is important to revert these values.

src/core/domain/useCases/validators/NewResourceValidator.ts
```diff
@@ -0,0 +1,6 @@
+import { ResourceValidationError } from './errors/ResourceValidationError';
+
+export interface NewResourceValidator {
+  // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+  validate(...args: any[]): Promise<void | ResourceValidationError>;
+}
```
MellyGray, line 6: If this is a generic interface, you can avoid the any type by using the T placeholder that can be specified when you implement the interface `NewResourceValidator<NewDataset>`
```suggestion
export interface NewResourceValidator<T> {
  validate(...args: T[]): Promise<void | ResourceValidationError>;
}
```

src/datasets/domain/models/NewDataset.ts
```diff
@@ -0,0 +1,21 @@
+import { DatasetLicense } from './Dataset';
+
+export interface NewDataset {
```
MellyGray: Isn't NewDataset more like a DTO since It appears to be merely a means of passing the necessary data to the 'create Dataset' endpoint? The real application data is the Dataset model, this interface is just for data transfer purposes 

Therefore, I would suggest naming it NewDatasetDTO and placing it nearer to the use case that utilizes it, rather than in the dataset model folder

src/datasets/domain/useCases/validators/NewDatasetValidator.ts
```diff
@@ -0,0 +1,228 @@
...
+export interface NewDatasetMetadataFieldAndValueInfo {
+  metadataFieldInfo: MetadataFieldInfo;
+  metadataFieldKey: string;
+  metadataFieldValue: NewDatasetMetadataFieldValue;
+  metadataBlockName: string;
+  metadataParentFieldKey?: string;
+  metadataFieldPosition?: number;
+}
+
+export class NewDatasetValidator implements NewResourceValidator {
```
MellyGray: I would suggest splitting the NewDatasetValidator class into multiple validator classes, such as MultipleMetadataFieldValidator, SingleMetadataFieldValidator, and so on. This is because following all the logic in such a large class can be difficult

src/datasets/infra/repositories/DatasetsRepository.ts
```diff
@@ -106,4 +110,25 @@ export class DatasetsRepository extends ApiRepository implements IDatasetsReposi
...
+      `/dataverses/${collectionId}/datasets`,
+      transformNewDatasetModelToRequestPayload(newDataset, datasetMetadataBlocks),
+    )
+      .then((response) => {
+        const responseData = response.data.data;
+        return {
+          persistentId: responseData.persistentId,
+          numericId: responseData.id,
+        };
+      })
```
MellyGray, line 129: In other repository method implementations this part was inside the transformer:

```
const responseData = response.data.data;
         return {
           persistentId: responseData.persistentId,
           numericId: responseData.id,
         };
```

Why is it outside of the repository in this case?

src/datasets/domain/useCases/validators/NewDatasetValidator.ts
```diff
@@ -0,0 +1,228 @@
...
+  async validate(resource: NewDataset, metadataBlocks: MetadataBlock[]): Promise<void | ResourceValidationError> {
```
MellyGray: I don't understand why this is an async method if it doesn't make any asynchronous calls and doesn't return anything. It seems to me as a completely synchronous method

We could be calling the validate method synchronously in the use case

```
async execute(newDataset: NewDataset, collectionId: string = 'root'): Promise<CreatedDatasetIdentifiers> {
    const metadataBlocks = await this.getNewDatasetMetadataBlocks(newDataset);

    this.newDatasetValidator.validate(newDataset, metadataBlocks);
    
    return this.datasetsRepository.createDataset(newDataset, metadataBlocks, collectionId);
}
```

MellyGray (changes requested):
Looks good. The use case is straightforward for the consumer, and it properly validates the dynamic metadata fields, raising specific errors

However, I believe that the validation logic could be refactored by splitting it into several classes to enhance readability

src/datasets/infra/repositories/DatasetsRepository.ts
GPortas: For simplicity, since the transformation is straightforward, I decided to do it directly without an intermediate transformer.

src/core/domain/useCases/validators/NewResourceValidator.ts
GPortas: I initially approached it precisely as you describe, since in the initial design I only considered the NewDataset parameter in the validator.

In a later step, as I also needed the MetadataBlocks, I had to remove the generic type, since the validator receives two parameters of different types: `resource: NewDataset, metadataBlocks: MetadataBlock[]`

Furthermore, it is possible that future validators will require different number parameters to carry out validations, so I find it appropriate to leave it open (with `any[]`) for the moment, just as we do with the UseCase base class params.

src/datasets/domain/models/NewDataset.ts
GPortas: I agree to rename it as you describe, using the DTO convention.

But I wouldn't place any DTO closer to the use case since it is also used in the repository. In my opinion the best way to locate them is equidistant from the use cases and repositories.

So I created a new folder called "dtos" at the same level as "useCases" and "repositories".

src/datasets/domain/useCases/validators/NewDatasetValidator.ts
GPortas: I agree. I also pushed a couple of fixes related to `getNewDatasetMetadataBlocks` method.

src/datasets/domain/useCases/validators/NewDatasetValidator.ts
GPortas: Refactored!

test/integration/environment/.env
GPortas: Reverted, since https://github.com/IQSS/dataverse/pull/10225 is merged.

src/datasets/domain/useCases/validators/BaseMetadataFieldValidator.ts
```diff
@@ -0,0 +1,101 @@
...
+export interface NewDatasetMetadataFieldAndValueInfo {
+  metadataFieldInfo: MetadataFieldInfo;
+  metadataFieldKey: string;
+  metadataFieldValue: NewDatasetMetadataFieldValueDTO;
+  metadataBlockName: string;
+  metadataParentFieldKey?: string;
+  metadataFieldPosition?: number;
+}
+
+export abstract class BaseMetadataFieldValidator {
```
MellyGray, line 14: It's challenging to discern which class is invoking which. I recommend several changes to enhance readability:

1. **Simplify the Base Class**: Retain only the `validate` and `createValidationError` methods in the base class. These methods will be accessible to all child validators. Move the rest of the specific single-field validation methods to the `SingleMetadataFieldValidator` class.

   ```typescript
   export abstract class BaseMetadataFieldValidator {
     abstract validate(newDatasetMetadataFieldAndValueInfo: NewDatasetMetadataFieldAndValueInfo): void;
     protected createValidationError(): FieldValidationError {
       // Implementation
     }
   }
   ```

2. **Refine the Single and Multiple Validator Classes**: Maintain both the single and multiple validator classes. However, the `MultipleMetadataFieldValidator` should now use an instance of `SingleMetadataFieldValidator` to invoke single-field validations, rather than calling methods from the base class.

   ```typescript
   export class MultipleMetadataFieldValidator extends BaseMetadataFieldValidator {
     constructor(private singleValidator: SingleMetadataFieldValidator) {
       super();
     }

     // Implementation of validate method
   }
   ```

3. **Structure the `MetadataFieldValidator` Class**: Ensure that `MetadataFieldValidator` contains instances of both the single and multiple field validators, enabling it to choose between them as needed.

   ```typescript
   export class MetadataFieldValidator extends BaseMetadataFieldValidator {
     constructor(private singleValidator: SingleMetadataFieldValidator, private multipleValidator: MultipleMetadataFieldValidator) {
       super();
     }

     // Implementation of validate method
   }
   ```
4. I would also remove `executeMetadataFieldValidator` method because it seems unnecessary, as child validation classes like `MultipleMetadataFieldValidator` can simply use `this.singleValidator.validate()` or `this.multipleValidator.validate()`

With these adjustments, we're adopting dependency injection instead of creating dependencies with `new`.

src/datasets/infra/repositories/DatasetsRepository.ts
MellyGray: I didn't mean adding another transformer, but returning this

{
          persistentId: responseData.persistentId,
          numericId: responseData.id,
        }

directly from the existing transformNewDatasetModelToRequestPayload transformer, just as in `transformDatasetPreviewsResponseToDatasetPreviewSubset` which I review recently and it was directly returning the expected response

src/datasets/domain/useCases/validators/NewDatasetValidator.ts
MellyGray: Thanks for applying the refactor, I left a few suggestions to improve the readability

src/datasets/domain/useCases/validators/BaseMetadataFieldValidator.ts
GPortas: I have updated the implementation following this approach.

I also renamed `NewDatasetValidator` to `NewDatasetResourceValidator` to better differentiate the implementation of `NewResourceValidator` from the child classes of `BaseMetadataFieldValidator`.

MellyGray (approved):
Looks good, thanks for applying the suggested changes

ekraffmiller:
looks good, all tests passed

Issue #81, opened 2023-08-09
Add an isolated functional testing mechanism


GPortas:
## Overview of the feature request

Add a functional testing mechanism for the use cases of the package that helps in the review and QA process of the PRs, so that the use cases can be tested in isolation, without having to depend on other repositories like the dataverse-frontend to prove that the use cases consumption works from a functional perspective.

These tests would help to verify that use cases can be imported with no issues, as stated in this comment: https://github.com/IQSS/dataverse-client-javascript/pull/116#issuecomment-1907711330

## What inspired the request?

Retrospective action item.

## What existing behavior do you want changed?

None

## Any brand new behavior do you want to add to js-dataverse?

None

## Any open or closed issues related to this feature request?

None

PR #116, opened 2024-01-17
109 - Add totalFilesCount to GetDatasetFiles use case

### LINKED ISSUE #109: Add total file count to GetDatasetFiles use case

As for the Dataset previews, the GetDatasetFiles should be extended to return the total file count, which is necessary for pagination.

We should follow the same implementation as for DatasetPreviewSubset.

This feature depends on: https://github.com/IQSS/dataverse/issues/10202

- Use case consumer

- https://github.com/IQSS/dataverse-client-javascript/pull/107#discussion_r1434858529

GetDatasetFiles use case to include total file count

None

- https://github.com/IQSS/dataverse/issues/10202

MellyGray:
This PR introduces `totalFilesCount` to the `GetDatasetFiles` use case, which is essential for pagination. This ensures consistency with the pagination implementation, as seen in `DatasetPreviewSubset`.

- Closes #109 

## Related Dataverse PRs:

- Depends on https://github.com/IQSS/dataverse/issues/10202

The implementation should be consistent with GetAllDatasetPreviews implementation, so just compare the code with that use case

Run integration tests or review GitHub action execution


GPortas (changes requested):
Looks good.

We need also to export the `FilesSubset` model in the `index.ts` file of the files folder, [just as we do with the `DatasetPreviewSubset`](https://github.com/IQSS/dataverse-client-javascript/blob/develop/src/datasets/index.ts#L47)

src/files/infra/repositories/transformers/fileTransformers.ts
```diff
@@ -1,14 +1,21 @@
...
-  const files: File[] = [];
+export const transformFilesResponseToFilesSubset = (
+    response: AxiosResponse,
+): FilesSubset => {
   const filesPayload = response.data.data;
+  const files: File[] = [];
   // eslint-disable-next-line  @typescript-eslint/no-explicit-any
-  filesPayload.forEach(function (filePayload: any) {
-    files.push(transformFilePayloadToFile(filePayload));
+  filesPayload.forEach(function (filesPayload: any) {
```
GPortas: Why have you changed the name of the variable inside the function from singular to plural?

Now we have `transformFilePayloadToFile(filesPayload)` which makes it a little confusing.

MellyGray:
> Looks good.
> 
> We need also to export the `FilesSubset` model in the `index.ts` file of the files folder, [just as we do with the `DatasetPreviewSubset`](https://github.com/IQSS/dataverse-client-javascript/blob/develop/src/datasets/index.ts#L47)

I addressed the change. But I'm worried about what would had happened if you hadn't noticed. I think that somehow we need some tests to test the package from the outside. To force the check on these index exports, and  also ensure that the package functions properly when used as a library

GPortas:
> > Looks good.
> > We need also to export the `FilesSubset` model in the `index.ts` file of the files folder, [just as we do with the `DatasetPreviewSubset`](https://github.com/IQSS/dataverse-client-javascript/blob/develop/src/datasets/index.ts#L47)
> 
> I addressed the change. But I'm worried about what would had happened if you hadn't noticed. I think that somehow we need some tests to test the package from the outside. To force the check on these index exports, and also ensure that the package functions properly when used as a library

Yes. There is an open issue related to functional tests that would be useful for what you mention: https://github.com/IQSS/dataverse-client-javascript/issues/81

I just updated the issue to link to this comment.

GPortas (commented):
Sorry, I forgot to submit this review.

Can you check the comment?

src/files/infra/repositories/transformers/fileTransformers.ts
MellyGray: My bad, I committed the change to restore the name

GPortas (approved):

PR #129, opened 2024-02-27
Adds returnDatasetVersion optional param to getFile use case

GPortas:

Adds returnDatasetVersion optional param to getFile use case

- None

Visual inspection and GitHub action run check

None


docs/useCases.md
```diff
@@ -310,6 +310,8 @@ The use case returns a [CreatedDatasetIdentifiers](../src/datasets/domain/models
 
 Returns a [File](../src/files/domain/models/File.ts) instance, given the search parameters to identify it.
 
+The use case will only return a [File](../src/files/domain/models/File.ts) object if the `returnDatasetVersion` optional parameter is not set or is false. If `returnDatasetVersion` is true, the use case will return a tuple of [File](../src/files/domain/models/File.ts) and [Dataset](../src/datasets/domain/models/Dataset.ts) objects (`[File, Dataset]`). The returned dataset object corresponds to the dataset version associated with the requested file.
```
MellyGray: From the user’s perspective, the handling of the return type can be confusing, as sometimes it is a File object and other times it is a tuple. I would suggest always returning a tuple, either [File, undefined] or [File, Dataset], to maintain consistency.

Alternatively, and this is my preferred approach, I would recommend splitting the functionality into two distinct use cases: getFile and getFileDatasetVersion. This adheres to the Single Responsibility Principle, ensuring each use case is responsible for only one thing.

Let me know what you think

MellyGray (changes requested):

MellyGray (approved):
LGTM!

PR #132, opened 2024-03-06
Fix CreateDataset use case

GPortas:

This PR fixes the Create Dataset use case, which was behaving incorrectly when integrated into the SPA.

- N/A

## Related Dataverse PRs:

- None

Added functional tests to help fix bugs found in the CreateDataset use case. This new type of test will be useful for testing the behavior of the use case as a package consumer. Currently only present for the CreateDataset use case.

In addition, the containerized environment has been fixed to support recently changes related to PIDs in the Dataverse application.

Visual inspection and GitHub action review

No

## Additional documentation:

None


src/metadataBlocks/infra/repositories/transformers/metadataBlockTransformers.ts
```diff
@@ -51,3 +56,28 @@ const transformPayloadMetadataFieldInfo = (
...
+  Object.keys(metadataFields).forEach((metadataFieldKey) => {
+    if (
+      metadataFields[metadataFieldKey].childMetadataFields &&
+      metadataFields[metadataFieldKey].childMetadataFields[newMetadataFieldKey]
+    ) {
+      isPresent = true
+    }
+  })
+  return isPresent
+}
```
MellyGray: While it might not make a significant difference, I noticed something that could easily improve the performance of these loops

I was thinking that if we collect all keys that exist as child fields into a set before the main loop begins, we could avoid some unnecessary iterations.

1. First, collect all `childFieldKeys`:

```javascript
const childFieldKeys = new Set();
Object.values(metadataBlockFieldsPayload).forEach((fieldInfo) => {
  if (fieldInfo.childFields) {
    Object.keys(fieldInfo.childFields).forEach(childKey => {
      childFieldKeys.add(childKey);
    });
  }
});
```

2. Then, execute the main loop:

```javascript
Object.keys(metadataBlockFieldsPayload).forEach((metadataFieldKey) => {
  if (!childFieldKeys.has(metadataFieldKey)) { // Check against the precomputed set for efficiency
    const metadataFieldInfoPayload = metadataBlockFieldsPayload[metadataFieldKey];
    metadataFields[metadataFieldKey] = transformPayloadMetadataFieldInfo(metadataFieldInfoPayload);
  }
});
```

Originally, we were checking if a field is a child field by iterating over all fields, which has a time complexity of O(n²). By storing the `childKeys` in a set, we reduce the time complexity of this operation to O(1), thereby improving efficiency

src/metadataBlocks/infra/repositories/transformers/metadataBlockTransformers.ts
```diff
@@ -51,3 +56,28 @@ const transformPayloadMetadataFieldInfo = (
...
+
+/**
+ * This method checks if a new metadata field key is already present in the metadata fields.
+ * We need this method since child fields are returned as sub-objects and as root objects in the response payload, so we don't want to replicate this behavior in the model.
+ *
+ * @param {Record<string, MetadataFieldInfo>} [metadataFields] - The current transformed metadata fields
+ * @param {string} newMetadataFieldKey - The new metadata field key to transform
+ *
+ * @returns {boolean}
+ */
```
MellyGray: Why did you add a comment to this function when the other transformer functions do not have any? The function is not even an exported function of the package

test/functional/datasets/CreateDataset.test.ts
```diff
@@ -0,0 +1,206 @@
...
+describe('execute', () => {
+  beforeEach(async () => {
+    ApiConfig.init(
+      TestConstants.TEST_API_URL,
+      DataverseApiAuthMechanism.API_KEY,
+      process.env.TEST_API_KEY
+    )
+  })
+
+  test('should successfully create a new dataset when all required fields are sent', async () => {
```
MellyGray, line 16: The test should include some assertions, such as:

```
const createdDataset = await createDataset.execute(testNewDataset).catch(() => {
    throw new Error('Dataset should be created');
  });

  expect(createdDataset).not.toBeNull();
  expect(createdDataset).toHaveProperty(X);
```

MellyGray (commented):
The fix is working, I tested the use case in the SPA. I left a few comments about the code

src/metadataBlocks/infra/repositories/transformers/metadataBlockTransformers.ts
GPortas: I decided to add a method doc because the behavior of the endpoint (child fields returned both as root objects and sub objects) and the fact that we have to do this filtering is not very intuitive, so I think the doc would help any new dev to understand this better.

Anyway I have removed both the method and the doc, which I don't think is strictly necessary, with the latest changes.

src/metadataBlocks/infra/repositories/transformers/metadataBlockTransformers.ts
GPortas: Changed. Thank you!

MellyGray (approved):
Looks good! Thanks for applying the suggested changes. I also like that you added new actions for the different tests

GPortas (commented):
This is already tested and validated as it is being consumed by the SPA.

PR #143, opened 2024-04-09
Create/destroy datasets on each IT/functional run

### LINKED ISSUE #138: Create and destroy test datasets on each integration/functional test run

Instead of pre-creating all the test datasets used in the functional and integration test suites through the setup.js script, we will create and destroy the datasets on each test run to ensure indempotence and prevent the recent issues related to race conditions.

Package developer.

Frontend dev team discussion

GPortas:

This PR adds the creation and deletion of datasets and collections in functional and integration tests. Each test file independently manages the creation and deletion of collections and datasets, thus avoiding dependencies between test files and race condition issues.

- Closes #138 

Visual inspection of the code and GitHub actions.


test/testHelpers/collections/collectionHelper.ts
```diff
@@ -31,3 +39,20 @@ export const createCollectionPayload = (): CollectionPayload => {
   }
   return collectionPayload
 }
+
+export async function createCollectionViaApi(collectionAlias: string): Promise<void> {
```
MellyGray: It doesn't seem to be returning void, since the create endpoint when used in the repository it returns CollectionPayload. I think we should use CollectionPayload as the returned type

test/integration/collections/CollectionsRepository.test.ts
```diff
@@ -3,40 +3,59 @@ import { TestConstants } from '../../testHelpers/TestConstants'
...
+  beforeAll(async () => {
     ApiConfig.init(
       TestConstants.TEST_API_URL,
       DataverseApiAuthMechanism.API_KEY,
       process.env.TEST_API_KEY
     )
+    try {
+      await createCollectionViaApi(TestConstants.TEST_CREATED_COLLECTION_ALIAS_2).then(
+        // eslint-disable-next-line  @typescript-eslint/no-explicit-any
+        (response: any) => (testCollectionId = response.data.data.id)
```
MellyGray: I think it should be CollectionPayload type instead of any

test/integration/datasets/DatasetsRepository.test.ts
```diff
@@ -101,47 +190,68 @@ describe('DatasetsRepository', () => {
...
+          throw new Error('Tests beforeAll(): Error while creating test dataset')
+        }
+      })
+
+      afterAll(async () => {
+        try {
+          await deletePublishedDatasetViaApi(testDatasetIds.persistentId)
+        } catch (error) {
+          throw new Error('Tests afterAll(): Error while deleting test dataset')
+        }
```
MellyGray: The try-catch block that handles the 'Error while deleting test dataset' appears to be duplicated in several places. Should we move this try-catch logic inside the `deletePublishedDatasetViaApi` function? That way, if any error occurs, it throws the 'Error while deleting test dataset' message, and we don’t need to repeat the logic everywhere

test/integration/datasets/DatasetsRepository.test.ts
```diff
@@ -48,48 +55,130 @@ describe('DatasetsRepository', () => {
...
+    const deleteDatasets = async () => {
+      try {
+        await deleteUnpublishedDatasetViaApi(firstDatasetIds.numericId)
+        await deleteUnpublishedDatasetViaApi(secondDatasetIds.numericId)
+        await deleteUnpublishedDatasetViaApi(thirdDatasetIds.numericId)
+        await deleteUnpublishedDatasetViaApi(fourthDatasetIds.numericId)
+      } catch (error) {
+        throw new Error('Tests afterAll(): Error while deleting test datasets')
+      }
+    }
```
MellyGray: I suggest storing the dataset IDs in an array, instead of in separate variables. This way, we can create the datasets using a loop, with each array element representing a dataset ID in sequence (first, second, etc.).

Then, for the deleteDatasets function, you can simply pass the array of IDs and use a loop to delete the datasets, rather than calling the delete function individually for each one

MellyGray (changes requested):
Looks good. I hope this will fix the race condition

test/integration/datasets/DatasetsRepository.test.ts
GPortas: I refactored the code to move the try catch _deletePublishedDatasetViaApi_ block, as well as other blocks, to the helper class methods.

GPortas:
Thanks for your review. Great suggestions.

MellyGray (approved):
LGTM!

g-saracca:
All Github actions look good!

PR #145, opened 2024-04-16
140 transform html to markdown for Collection description

### LINKED ISSUE #140: For getCollection useCase, transform html to markdown
The description field of the Collection may contain html markup, so that needs to be converted to markdown, which is what the SPA expects.  
For example the JSF Collection page:
 
 
Screenshot 2024-04-02 at 12 42 48 PM

ekraffmiller:
Transforms html that may be in Collection description to markdown

- Closes #140 

## Related Dataverse PRs:

- Depends on #

Create a new Collection in JSF, with a description that contains html.  The getCollection useCase should return a collection that has a description with markdown.


test/testHelpers/collections/collectionHelper.ts
```diff
@@ -6,15 +6,15 @@ const COLLECTION_ID = 11111
...
+const COLLECTION_DESCRIPTION_HTML = 'This is an <b>example</b> collection used for testing.'
+const COLLECTION_DESCRIPTION_MARKDOWN = 'This is an **example** collection used for testing.'
 export const createCollectionModel = (): Collection => {
   const collectionModel: Collection = {
     id: COLLECTION_ID,
     alias: COLLECTION_ALIAS_STR,
     name: COLLECTION_NAME_STR,
     affiliation: COLLECTION_AFFILIATION_STR,
-    description: COLLECTION_DESCRIPTION_STR,
+    description: COLLECTION_DESCRIPTION_HTML,
```
MellyGray: Shouldn't this be the `COLLECTION_DESCRIPTION_MARKDOWN` since this is the transformed description?
```suggestion
    description: COLLECTION_DESCRIPTION_MARKDOWN,
```

test/testHelpers/collections/collectionHelper.ts
```diff
@@ -26,8 +26,16 @@ export const createCollectionPayload = (): CollectionPayload => {
...
     isPartOf: { type: DvObjectType.DATAVERSE, identifier: 'root', displayName: 'Root' }
   }
   return collectionPayload
 }
+export const transformedCollectionModel: CollectionPayload = {
```
MellyGray: think we could use the `createCollectionModel` instead of creating a new `transformedCollectionModel`, since all collection models should be transformed

test/unit/collections/CollectionsRepository.test.ts
```diff
@@ -45,7 +46,7 @@ describe('CollectionsRepository', () => {
         const actual = await sut.getCollection(testCollectionModel.id)
 
         expect(axios.get).toHaveBeenCalledWith(expectedApiEndpoint, expectedRequestConfigApiKey)
-        expect(actual).toStrictEqual(testCollectionModel)
+        expect(actual).toStrictEqual(transformedCollectionModel)
```
MellyGray: I think we should use the `createCollectionModel` for the assertion

MellyGray (changes requested):

test/testHelpers/collections/collectionHelper.ts
ekraffmiller: thanks, much cleaner to do it this way, we only need one model for the tests

MellyGray (approved):
LGMT!

GPortas (approved):

PR #147, opened 2024-04-22
139 add is released to get collection

### LINKED ISSUE #139: add isReleased to the getCollection useCase
For the SPA, we need to display an 'Unpublished' label on the Collections page, if the collection hasn't been released.  To do this we need to add an isReleased field to the payload of the getCollection useCase.

This issue depends on adding the field to the API: https://github.com/IQSS/dataverse/issues/10448

ekraffmiller:
in the getCollection useCase, sets the isReleased field in the Collection Payload, using the isReleased field in the API response.

- Closes #139 

## Related Dataverse PRs:

- Depends on #

Run the CollectionsRepository integration test


test/integration/files/FilesRepository.test.ts
```diff
@@ -87,7 +87,14 @@ describe('FilesRepository', () => {
...
       })
     // Registering test file 1
+
     await registerFileViaApi(uploadFileResponse.data.data.files[0].dataFile.id)
+      .then()
+      .catch((err) => {
+        console.log(err.message)
+        console.log(err.response.data)
+        throw new Error(`Tests beforeAll(): Error while registering file ${testTextFile1Name}`)
+      })
```
MellyGray: The then() is not necessary since it's not performing any action on the promise. We can also remove the console.logs
```suggestion
    await registerFileViaApi(uploadFileResponse.data.data.files[0].dataFile.id)
      .catch((err) => {
        throw new Error(`Tests beforeAll(): Error while registering file ${testTextFile1Name}`)
      })
```

MellyGray (changes requested):
Looks good! I've left a comment to remove some console logs

ekraffmiller:
Thanks @MellyGray , I made your suggested changes...also I made the change that's described here: https://github.com/IQSS/dataverse-client-javascript/pull/150 so that the tests would succeed.  I think if this PR is merged, we can just close the other one

test/integration/files/FilesRepository.test.ts
ekraffmiller: thanks @MellyGray,I made the suggested changes

MellyGray (approved):
LGTM!

PR #153, opened 2024-05-09
Adds use cases for uploading files through direct upload API

### LINKED ISSUE #152: Create use case for uploading files via Direct Upload / S3

The goal is to create a use case for uploading files using Direct DataFile Upload/Replace API: https://guides.dataverse.org/en/latest/developers/s3-direct-upload-api.html#to-add-multiple-uploaded-files-to-the-dataset

The approach to follow will allow the steps for direct upload to be encapsulated in a use case, so that this mechanism can be directly invoked by the SPA or another consumer.

Possibly this issue may require changes to the API.

This use case will be consumed from the Dataverse SPA (File upload page), although it should be applicable to other scenarios.

- Use case consumer

- Q2 plan: https://docs.google.com/document/d/1uE0s-XxL-cLzcXJ3buyrbpc95bqKIewagaOtVAsMPGI

None

Use case for uploading files via Direct Upload / S3

- None

GPortas:

This PR includes the use cases to support file uploads through the Direct Upload API. These use cases are entirely based on the Direct Upload API, so for the review of this PR, I recommend to rely on: https://guides.dataverse.org/en/latest/developers/s3-direct-upload-api.html

- Closes #152

## Related PRs:

- https://github.com/IQSS/dataverse-frontend/pull/397

This PR does not incorporate use case functional tests because the tests are run outside the container network and we encountered issues with hostnames not being recognized for the generated destination URLs. However, coverage is still high, over 95%, and the different scenarios are tested with integration and unit tests.

Visual inspection of the code and unit and integration tests execution.

## Additional documentation:

- https://guides.dataverse.org/en/latest/developers/s3-direct-upload-api.html


package.json
```diff
@@ -47,6 +47,7 @@
     "@types/jest": "^29.5.12",
     "@typescript-eslint/eslint-plugin": "5.51.0",
     "@typescript-eslint/parser": "5.51.0",
+    "@web-std/file": "3.0.3",
```
g-saracca, line 50: What is the difference from using this package Blob and File interfaces than using the native ones?

src/files/domain/repositories/IFilesRepository.ts
```diff
@@ -4,8 +4,10 @@ import { FileUserPermissions } from '../models/FileUserPermissions'
 import { FileSearchCriteria, FileOrderCriteria } from '../models/FileCriteria'
 import { FileCounts } from '../models/FileCounts'
 import { FileDownloadSizeMode } from '../models/FileDownloadSizeMode'
-import { File } from '../models/File'
+import { File as FileModel } from '../models/File'
```
g-saracca: Just an idea: 
What if we rename our own File interface to something like IFile or FileModel, so it doesn't confuse with the native File interface and we don't need to rename it on the imports.

src/files/infra/clients/DirectUploadClient.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  public async uploadFile(
+    datasetId: number | string,
+    file: File,
+    progress: (now: number) => void,
+    abortController: AbortController,
+    destination?: FileUploadDestination
+  ): Promise<string> {
+    if (destination == undefined) {
```
g-saracca: Maybe strict equality here? `if (destination === undefined)`

src/files/infra/clients/DirectUploadClient.ts
```diff
@@ -0,0 +1,220 @@
...
+  ): Promise<void> {
+    try {
+      const arrayBuffer = await file.arrayBuffer()
+      await axios.put(destination.urls[0], arrayBuffer, {
+        headers: {
+          'Content-Type': 'application/octet-stream',
+          'Content-Length': file.size.toString(),
+          'x-amz-tagging': 'dv-state=temp'
+        },
+        timeout: 60000,
```
g-saracca: 60000 could be in its own constant and with [numeric separator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#numeric_separators) to improve readability

src/files/infra/clients/DirectUploadClient.ts
```diff
@@ -0,0 +1,220 @@
...
+            'Content-Length': fileSlice.size
+          },
+          maxBodyLength: Infinity,
+          maxContentLength: Infinity,
+          timeout: 60000,
+          signal: abortController.signal
+        })
+        const eTag = response.headers['etag'].replace(/"/g, '')
+        eTags[`${index + 1}`] = eTag
+        progress(this.progressAfterUrlGeneration + progressPartSize * (index + 1))
```
g-saracca: Nice implementation for the progress!! just one thing, sometimes the progress can be a large decimal number like `67.14285714285714`, maybe you can Math.round it to always send a rounded number?

src/files/infra/repositories/transformers/fileUploadDestinationsTransformers.ts
```diff
@@ -0,0 +1,50 @@
...
+  complete?: string
+  abort?: string
+}
+
+export const transformUploadDestinationsResponseToUploadDestination = (
+  response: AxiosResponse
+): FileUploadDestination => {
+  const fileUploadDestinationsPayload = response.data.data
+
+  if (fileUploadDestinationsPayload.url != undefined) {
```
g-saracca: strict equality here also?

src/files/infra/repositories/FilesRepository.ts
```diff
@@ -202,6 +206,52 @@ export class FilesRepository extends ApiRepository implements IFilesRepository {
...
+      .then((response) => transformUploadDestinationsResponseToUploadDestination(response))
+      .catch((error) => {
+        throw error
+      })
+  }
+
+  public async addUploadedFileToDataset(
+    datasetId: number | string,
+    uploadedFileDTO: UploadedFileDTO
+  ): Promise<undefined> {
```
g-saracca, line 229: Could be helpful for the consumer of the use case to receive something different that undefined here if the operation suceeds ? what will be the final output that the use case will send?

src/files/infra/clients/DirectUploadClient.ts
```diff
@@ -0,0 +1,220 @@
...
+    this.maxMultipartRetries = maxMultipartRetries
+  }
+
+  public async uploadFile(
+    datasetId: number | string,
+    file: File,
+    progress: (now: number) => void,
+    abortController: AbortController,
+    destination?: FileUploadDestination
+  ): Promise<string> {
```
g-saracca, line 42: What about returning an object like `Promise<{ storageId : string }>`, perhaps is more descriptive

g-saracca (changes requested):
First of all congratulations for such a great implementation and research about file uploading! 👏🏼 
Overall the code looks very good and solid, I've left a few suggestions for minor improvements.

package.json
GPortas: When running JavaScript tests in Node.js, the File and Blob APIs are not natively available since we are not running on a browser. 

@web-std/file is a library that can be used to import and polyfill File and Blob objects in environments where they are not natively supported.

src/files/infra/repositories/FilesRepository.ts
GPortas: This method is opaque to the use case (In fact, the use case calling this method returns `Promise<void>`). I see no value in changing the return type. In repository calls whose response is not parsed to return anything, we always return undefined.

src/files/infra/clients/DirectUploadClient.ts
GPortas: Personally, I don't like very much the idea of making the returned data structure more complex in favor of improving readability here, especially when we provide documentation of the use case to clearly understand which data is returned as a string.

g-saracca: yep, sounds good, lets keep the string 👍🏼

g-saracca (approved):
LGTM! Thanks!

ekraffmiller:
Looks good and all tests pass, merging!

PR #185, opened 2024-09-02
GetCollectionItems use case search text filtering

### LINKED ISSUE #183: Extend GetCollectionItems to support search text filtering
## Overview of the Feature Request

Extend GetCollectionItems to support search text filtering, since it is going to be required for the collection page and the search home page.

## What Inspired the Request?

- New search home page

## What Existing Behavior Do You Want Changed?

- None

## Any Brand New Behavior Do You Want to Add to js-dataverse?

- Search text filtering in GetCollectionItems use case

## Any Open or Closed Issues Related to This Feature Request?

- https://github.com/IQSS/dataverse-client-javascript/pull/181

GPortas:

Adds search text filtering option to GetCollectionItems use case.

⚠️ Environment variables should be changed to unstable image tag before merging into develop. https://github.com/IQSS/dataverse/pull/10811 should be merged before applying this change.

- Closes #183 

## Related Dataverse PRs:

- https://github.com/IQSS/dataverse/pull/10811

I have reduced branch test coverage from 95% to 90%.

The pre-commit hook was failing because branch coverage was not being reached in classes where there should be branch coverage, and it was showing 0%. I have created an issue to address this problem separately: https://github.com/IQSS/dataverse-client-javascript/issues/186

Visual inspection of code and github action executions


src/collections/infra/repositories/CollectionsRepository.ts
```diff
@@ -143,10 +148,22 @@ export class CollectionsRepository extends ApiRepository implements ICollections
     if (offset !== undefined) {
       queryParams.start = offset
     }
-    return this.doGet('/search?q=*&sort=date&order=desc', true, queryParams)
+    if (collectionSearchCriteria !== undefined) {
+      this.applyCollectionSearchCriteriaToQueryParams(queryParams, collectionSearchCriteria)
```
g-saracca, line 152: Shouldn't we check here also that `collectionSearchCritera` has some keys in it?
Something like I believe: `Object.keys(collectionSearchCriteria).length > 0,`, or even also checking first is an object and not an array object, just in case an empty object is send as `collectionSearchCriteria` which won't be undefined.
I understand that if the js-dataverse library is consumed by a Typescript project this shouldn't be a problem, but it could be a problem used from a javascript only project.

g-saracca (commented):
Looks good, just one comment about an if check, maybe we don't need.

src/collections/infra/repositories/CollectionsRepository.ts
GPortas: @g-saracca I think the check you mention is a general problem for the whole package for JavaScript consumers. If we perform this check at this point, I believe it would be necessary to do it in many other parts of the package code.

I think it's the responsibility of the consumer, who chooses to use JavaScript with the associated risks of dynamic typing, to ensure they are sending the correct types.

GPortas: And the docs, (useCases.md), is useful for consumers to know what types should be used.

g-saracca: Yes, I think you are right.
I think if we start getting feedback or comments from javascript only users we might think of something at the moment, but for now it's ok.

g-saracca (approved):
Looks good!

PR #192, opened 2024-09-18
fix: change the data type of isReleased to boolean

### LINKED ISSUE #189: Handle isReleased property as boolean
In the `getCollection` use case we are treating the `isReleased` property as a string but the API response returns a boolean.
We need to change the `Collection` and `CollectionPayload` type interfaces and update any test if needed.

After this work is done we should remove the hardcoded value from the frontend repo in [this part of the code](https://github.com/IQSS/dataverse-frontend/blob/develop/src/collection/infrastructure/mappers/JSCollectionMapper.ts#L17).
The `isReleased` property is correctly treated as a boolean in the frontend repo.

ChengShi-1:

The **isReleased** is supposed to be boolean, instead of string data type, in order to align with API responses.

- Closes #189 

## Related Dataverse PRs:

- Depends on #

After this work is done we should remove the hardcoded value from the frontend repo in [this part of the code](https://github.com/IQSS/dataverse-frontend/blob/develop/src/collection/infrastructure/mappers/JSCollectionMapper.ts#L17).

The isReleased property is correctly treated as a boolean in the frontend repo.

No

### DIFF
--- a/src/collections/domain/models/Collection.ts
+++ b/src/collections/domain/models/Collection.ts
@@ -4,7 +4,7 @@ export interface Collection {
   id: number
   alias: string
   name: string
-  isReleased: string
+  isReleased: boolean
   affiliation?: string
   description?: string
   isPartOf: DvObjectOwnerNode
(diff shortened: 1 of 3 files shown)


ekraffmiller (approved):
Looks good, approved!

GPortas (approved):

PR #193, opened 2024-09-19
Api token management use cases

### LINKED ISSUE #190: Create use cases for API Token management
## Overview of the Feature Request

Create use cases for API Token management:

- Get api token
- Create api token
- Recreate api token
- Revoke api token

These use cases will be consumed by the SPA.

## What Inspired the Request?

- https://github.com/IQSS/dataverse-frontend/pull/479 

## What Existing Behavior Do You Want Changed?

- None

## Any Brand New Behavior Do You Want to Add to js-dataverse?

- Create use cases for API Token management

GPortas:

Adds API token management use cases.

Remember to update the .env variables to use `unstable` before merging

- Closes #190 

## Related Dataverse PRs:

- Depends on https://github.com/IQSS/dataverse/pull/10858

Visual inspection of code and github actions.


docs/useCases.md
```diff
@@ -52,6 +52,10 @@ The different use cases currently available in the package are classified below,
 - [Users](#Users)
   - [Users read use cases](#users-read-use-cases)
     - [Get Current Authenticated User](#get-current-authenticated-user)
+    - [Get Current API Token](#get-current-api-token)
+  - [Users write use cases](#users-write-use-cases)
+    - [Delete Current API Token](#delete-current-api-token)
+    - [Recreate Current API Token](#recreate-current-api-token)
```
g-saracca, line 58: I think we arre missing one use case for **creating** an API Token, or maybe should we use the Recreate use case for that also?

src/users/infra/repositories/transformers/apiTokenInfoTransformers.ts
```diff
@@ -0,0 +1,26 @@
+import { AxiosResponse } from 'axios'
+import { ApiTokenInfo } from '../../../domain/models/ApiTokenInfo'
+
+export const transformGetApiTokenResponseToApiTokenInfo = (
+  response: AxiosResponse
+): ApiTokenInfo => {
+  const apiTokenInfoPayload = response.data.data
+  const messageParts = apiTokenInfoPayload.message.split(' ')
+  const expirationDateFormattedTimestamp = `${messageParts[4]}T${messageParts[5]}`
```
g-saracca, line 9: This is ok but in the future we could modify the API to return a JSON object with two properties - `token` and `expirationDate` instead of just a string which makes it more difficult to obtain the response data.

docs/useCases.md
GPortas: Yes, the same use case is used when creating an API Token and there is not an existing one.

src/users/infra/repositories/transformers/apiTokenInfoTransformers.ts
GPortas: I totally agree, and mentioned what you suggest on the API extension PR: https://github.com/IQSS/dataverse/pull/10858

I've kept the string messages as responses for backward compatibility.

g-saracca (approved):

ChengShi-1 (approved):
Great. Approve

PR #201, opened 2024-10-11
Experimental bearer token auth mechanism

GPortas:

- Closes #

## Related Dataverse PRs:

- Depends on #

### DIFF
--- a/src/core/infra/repositories/apiConfigBuilders.ts
+++ b/src/core/infra/repositories/apiConfigBuilders.ts
@@ -32,6 +32,13 @@ export const buildRequestConfig = (
         requestConfig.headers['X-Dataverse-Key'] = ApiConfig.dataverseApiKey
       }
       break
+    case DataverseApiAuthMechanism.BEARER_TOKEN:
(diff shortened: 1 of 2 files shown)


src/core/infra/repositories/apiConfigBuilders.ts
```diff
@@ -32,6 +32,15 @@ export const buildRequestConfig = (
         requestConfig.headers['X-Dataverse-Key'] = ApiConfig.dataverseApiKey
       }
       break
+    case DataverseApiAuthMechanism.BEARER_TOKEN:
+      // TODO: ROCP_token name should be configurable and extracted to an environment variable / application constant.
+      let token = localStorage.getItem('ROCP_token')
+      if (token) {
+        token = token.replace(/^"(.*)"$/, '$1')
```
g-saracca, line 39: @GPortas we could use `JSON.parse(token)` to avoid this regex replacement.
Something like:
```
const token = localStorage.getItem('ROCP_token') ? JSON.parse(localStorage.getItem('ROCP_token')) : null
```
And then in the `if` condition we should also check that the parsed token is not an empty string I think.

PR #207, opened 2024-10-28
Use  ':root', rather than alias 'root' for default root collection

### LINKED ISSUE #191: Support dynamic alias for the root collection
Currently the default value of collection alias in  the Collection use cases is 'root'.  This can cause a problem if the alias of the root collection is edited.  To manage this the Dataverse API has a keyword, ':root', to get the root collection.  

Need to analyze how to update the default collection alias being used in the Collection Use cases.  For example, it makes sense to use ':root' as a default value for Get Collection, but what do we want to use as the default value for Create Collection?  

See SPA issue: https://github.com/IQSS/dataverse-frontend/issues/489

ekraffmiller:
The root collection alias, 'root' is not a constant, it is an editable field in the Collection. So we need to use the special keyword ':root' to get the root collection.  This affects all use cases that default to the root collection if no collectionId param is provided.

- Closes #191 

## Related Dataverse PRs:

- Depends on #

Visual inspection of the code, including the updated Readme for the use cases, and review of the tests that use the default root collection.

### DIFF
--- a/src/collections/domain/useCases/GetCollection.ts
+++ b/src/collections/domain/useCases/GetCollection.ts
@@ -13,7 +13,7 @@ export class GetCollection implements UseCase<Collection> {
    * Returns a Collection instance, given the search parameters to identify it.
    *
    * @param {number | string} [collectionIdOrAlias = ':root'] - A generic collection identifier, which can be either a string (for queries by CollectionAlias), or a number (for queries by CollectionId)
-   * If this parameter is not set, the default value is: 'root'
+   * If this parameter is not set, the default value is: ':root'
    * @returns {Promise<Collection>}
    */
   async execute(collectionIdOrAlias: number | string = ROOT_COLLECTION_ID): Promise<Collection> {
--- a/src/collections/domain/useCases/GetCollection.ts
+++ b/src/collections/domain/useCases/GetCollection.ts
@@ -1,6 +1,6 @@
 import { UseCase } from '../../../core/domain/useCases/UseCase'
 import { ICollectionsRepository } from '../repositories/ICollectionsRepository'
-import { Collection, ROOT_COLLECTION_KEY } from '../models/Collection'
+import { Collection, ROOT_COLLECTION_ID } from '../models/Collection'
 
 export class GetCollection implements UseCase<Collection> {
   private collectionsRepository: ICollectionsRepository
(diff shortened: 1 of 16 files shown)


src/collections/domain/useCases/GetCollection.ts
```diff
@@ -12,11 +12,11 @@ export class GetCollection implements UseCase<Collection> {
   /**
    * Returns a Collection instance, given the search parameters to identify it.
    *
-   * @param {number | string} [collectionIdOrAlias = 'root'] - A generic collection identifier, which can be either a string (for queries by CollectionAlias), or a number (for queries by CollectionId)
+   * @param {number | string} [collectionIdOrAlias = ':root'] - A generic collection identifier, which can be either a string (for queries by CollectionAlias), or a number (for queries by CollectionId)
    * If this parameter is not set, the default value is: 'root'
```
g-saracca:
```suggestion
   * If this parameter is not set, the default value is: ':root'
```

g-saracca (commented):
This looks really good @ekraffmiller! 
I suggested a commit (maybe is not needed) and have a doubt about this part of the code in `collectionHelper.ts`, should be use ':root' instead of 'root' also there?
```
export async function createCollectionViaApi(
  collectionAlias: string,
  parentCollectionAlias: string | undefined = undefined
): Promise<CollectionPayload> {
  try {
    if (parentCollectionAlias == undefined) {
      parentCollectionAlias = 'root'
    }
```

Another thing, just out of curiosity, do all API endpoints work with this special id ':root'?

ekraffmiller:
> Another thing, just out of curiosity, do all API endpoints work with this special id ':root'?



> This looks really good @ekraffmiller! I suggested a commit (maybe is not needed) and have a doubt about this part of the code in `collectionHelper.ts`, should be use ':root' instead of 'root' also there?
> 
> ```
> export async function createCollectionViaApi(
>   collectionAlias: string,
>   parentCollectionAlias: string | undefined = undefined
> ): Promise<CollectionPayload> {
>   try {
>     if (parentCollectionAlias == undefined) {
>       parentCollectionAlias = 'root'
>     }
> ```
> 
> Another thing, just out of curiosity, do all API endpoints work with this special id ':root'?

Yes, it should be `:root` thanks!
All of the endpoints should work with this id, as far as I know.  I left two use cases as is (not using :root).  For 'publishCollection', the root collection should always be published, so I didn't think it needed it.  For getCollectionItems, the API call doesn't require collectionId.

g-saracca (approved):

ofahimIQSS:
Did visual inspection of code. Merging PR

PR #230, opened 2024-12-09
Get Collection Items extension - facets - filter queries - sort/type

### LINKED ISSUE #229: Get Collection Items use case extension
As part of SPA Q4: Collections: all filtering options, we need to update getCollectionItems use case to:

### Add Facets results

For the SPA Q4 feature "Collection Page: all filtering options", we need to get the facet results for the Collection Page. 
Screenshot 2024-12-04 at 7 11 16 PM

To get this data, we can use the use `show_facets=true` param in the Dataverse Search API.  
The getCollectionItems use case should be updated to return this data, possibly by adding a boolean field `showFacets` to the CollectionSearchCriteria object.  

   
### Filter results by facets

This is currently supported in the Dataverse Seach API by using the param fq, for example `fq=publicationDate:2024`.  See more details in the [Dataverse Search API](https://guides.dataverse.org/en/latest/api/search.html#narrowed-to-collection-show-relevance-and-facets) documentation. 

The CollectionSearchCriteria passed to getCollectionItems use case can be updated to included the facet queries.
Screenshot 2024-12-04 at 8 10 04 PM

### Add sort order

For the SPA Q4 feature "Collection Page: all filtering options", we need to implement the Sort button on the collections page:
Screenshot 2024-12-04 at 6 43 42 PM

To support this, we need to update the CollectionSearchCriteria used in getCollectionItems use case, to include a sorting information.  There are two allowed sort types: 'name' and 'date'.  The sort order can also be specified as ascending or descending.  [See Dataverse Search API ](https://guides.dataverse.org/en/latest/api/search.html#parameters) or more details.

g-saracca:

Extends the Get Collection Items use case. 
Now returns the `facets` according to the search.
Also now accepts 3 more properties in the `CollectionSearchCriteria` - `sort`, `order` and `filterQueries`.
`sort`: 'name' or 'date'
`type:` 'asc' or 'desc'
`filterQueries`: an strings Array of type `string`:`string`

- Closes #229 

Visual code inspection and run tests.

No


src/collections/domain/models/CollectionSearchCriteria.ts
```diff
@@ -1,16 +1,38 @@
...
   ) {}
 
   withSearchText(searchText: string | undefined): CollectionSearchCriteria {
-    return new CollectionSearchCriteria(searchText, this.itemTypes)
+    return new CollectionSearchCriteria(searchText, this.itemTypes, this.sort, this.order)
   }
 
   withItemTypes(itemTypes: CollectionItemType[] | undefined): CollectionSearchCriteria {
-    return new CollectionSearchCriteria(this.searchText, itemTypes)
+    return new CollectionSearchCriteria(this.searchText, itemTypes, this.sort, this.order)
```
ekraffmiller: To be consistent it looks like think these calls need a filterQueries parameter
```suggestion
    return new CollectionSearchCriteria(this.searchText, itemTypes, this.sort, this.order, this.filterQueries)
```

src/collections/domain/models/CollectionSearchCriteria.ts
```diff
@@ -1,16 +1,38 @@
...
+  }
+
+  withSort(sort: SortType | undefined): CollectionSearchCriteria {
+    return new CollectionSearchCriteria(this.searchText, this.itemTypes, sort, this.order)
```
ekraffmiller:
```suggestion
    return new CollectionSearchCriteria(this.searchText, this.itemTypes, sort, this.order, this.filterQueries)
```

src/collections/domain/models/CollectionSearchCriteria.ts
```diff
@@ -1,16 +1,38 @@
...
+  }
+
+  withOrder(order: OrderType | undefined): CollectionSearchCriteria {
+    return new CollectionSearchCriteria(this.searchText, this.itemTypes, this.sort, order)
```
ekraffmiller:
```suggestion
    return new CollectionSearchCriteria(this.searchText, this.itemTypes, this.sort, order, this.filterQueries)
```

src/collections/domain/models/GetCollectionItemsQueryParams.ts
```diff
@@ -0,0 +1,23 @@
+export enum GetCollectionItemsQueryParams {
```
ekraffmiller: Since the GetCollectionItemsQueryParams enum is only used in CollectionRepository, and not in the interface or use cases, it seems to be implementation specific, and can stay in CollectionRepository.ts rather than moved here.  Then this file can be renamed to something like CollectionSearchCriteriaTypes?

src/collections/infra/repositories/transformers/collectionTransformers.ts
```diff
@@ -79,7 +83,10 @@ export const transformCollectionItemsResponseToCollectionItemSubset = (
 ): CollectionItemSubset => {
   const responseDataPayload = response.data.data
   const itemsPayload = responseDataPayload.items
+  const facets = responseDataPayload.facets as CollectionItemsFacetPayload
```
ekraffmiller: this is just cosmetic, but to be consistent with the other var names, it could be good to rename `facets`-> `facetsPayload`, and `transformedFacets` -> `facets`

ekraffmiller (commented):
Looks great, I just have some minor comments below

src/collections/domain/models/GetCollectionItemsQueryParams.ts
g-saracca: Thanks! I deleted this file and move GetCollectionItemsQueryParams to CollectionRepository.ts and move the other types directly into CollectionSearchCriteria.ts

g-saracca:
@ekraffmiller all changes addressed, thanks for the review!

ekraffmiller (approved):
looks good, approved!

ekraffmiller:
@g-saracca I realized there is one more change needed, there is a sort type "Relevance" used when the CollectionSearchCriteria contains search text. For the API, `sort=score`. For example: 

Screenshot 2024-12-17 at 8 01 21 AM


I'm going to put the PR back to InReview, sorry about that.

g-saracca:
Hi @ekraffmiller , in the api endpoint docs there is nothing about sort by relevance, there is `show_relevance` but it's for something different. Sort can be `date` or `name` and Order `asc` or `desc`, but nothing about relevance.
Perhaps we need to extend the api endpoint to achieve that but maybe we can continue without that and do it in another issue after the api extension is done.

https://guides.dataverse.org/en/latest/api/search.html

ekraffmiller:
Ah ok, so when the user chooses sort by relevance, that is the same as doing a text search without a sort option. Makes sense, I don't think we need to update the API, I was just confused by the url params in the frontend.

g-saracca:
Right, when you search something the url looks like this:
https://beta.dataverse.org/dataverse/root/?q=foo
And when clicking Sort By Releveance the url turns to this:
https://beta.dataverse.org/dataverse/root?q=foo&types=dataverses%3Adatasets%3Afiles&page=1&sort=score&order=desc

ofahimIQSS:
Tests are passing - Merging PR

PR #234, opened 2025-01-03
Update file restriction fields and tabular information to the file preview

### LINKED ISSUE #221: add file access(restricted) and permission canDownloadFile check to FilePreview
In order to show lock icon for restricted files in the collection page card, referring to this issue in the front-end repo https://github.com/IQSS/dataverse-frontend/issues/548, we need deliver access and permission information for each file to the UI, so we could check whether the lock icon is needed. In filePreview, the following properties are missing, and needed to be added.
`restricted: boolean`
`canDownloadFile: boolean`
`categories: array of string` "categories" would be similar to what it is in metadata api, for example, `{"label":"blob","description":"","restricted":true,"categories":["Data","Documentation"],"id":6}`

For tabular files, there are some properties needed additionally.
`tabularTags: array of string `for example,`{"tabularTags" : ["Event", "Genomics", "Geospatial"]}`
`variables: number/int` shows how many variables we have for the tabular file
`observations: number/int `shows how many observations for the tabular file

In dataset item preview, thumbnail waits to be added as well.  
`thumbnail: string`

Waiting for https://github.com/IQSS/dataverse/issues/11027

ChengShi-1:
In collection page, some missing information is needed for every card item, including the lock icon, tags(categories), and tabular data info, so we need update in FilePreview model.

New fields include:
```
 restricted: boolean
 canDownloadFile: boolean
 categories?: string[]
 tabularTags?: string[]
 variables?: number
 observations?: number

```

In datasetPreview, new field added
` image_url: string`

- Closes #221 
## Related Dataverse PRs:

- Depends on https://github.com/IQSS/dataverse/pull/11097

Run all testcases and visually inspect

### DIFF
--- a/test/integration/collections/CollectionsRepository.test.ts
+++ b/test/integration/collections/CollectionsRepository.test.ts
@@ -399,6 +399,7 @@ describe('CollectionsRepository', () => {
       )
       expect(actualDatasetPreview.parentCollectionName).toBe(expectedCollectionsName)
       expect(actualDatasetPreview.type).toBe(CollectionItemType.DATASET)
+      expect(actualDatasetPreview.imageUrl).toBe('http://dataverse.com')
 
       expect(actualCollectionPreview.name).toBe(expectedCollectionsName)
       expect(actualCollectionPreview.alias).toBe(testSubCollectionAlias)
--- a/test/integration/collections/CollectionsRepository.test.ts
+++ b/test/integration/collections/CollectionsRepository.test.ts
@@ -399,7 +399,6 @@ describe('CollectionsRepository', () => {
       )
       expect(actualDatasetPreview.parentCollectionName).toBe(expectedCollectionsName)
       expect(actualDatasetPreview.type).toBe(CollectionItemType.DATASET)
-      expect(actualDatasetPreview.imageUrl).toBe('http://dataverse.com')
 
       expect(actualCollectionPreview.name).toBe(expectedCollectionsName)
       expect(actualCollectionPreview.alias).toBe(testSubCollectionAlias)
--- a/test/integration/collections/CollectionsRepository.test.ts
+++ b/test/integration/collections/CollectionsRepository.test.ts
@@ -20,8 +20,12 @@ import {
   ROOT_COLLECTION_ALIAS
 } from '../../testHelpers/collections/collectionHelper'
 import { CollectionPayload } from '../../../src/collections/infra/repositories/transformers/CollectionPayload'
-import { uploadFileViaApi } from '../../testHelpers/files/filesHelper'
-import { deleteUnpublishedDatasetViaApi } from '../../testHelpers/datasets/datasetHelper'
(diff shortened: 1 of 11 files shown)


test/integration/collections/CollectionsRepository.test.ts
```diff
@@ -293,7 +293,7 @@ describe('CollectionsRepository', () => {
 
       const expectedFileMd5 = '68b22040025784da775f55cfcb6dee2e'
       const expectedDatasetCitationFragment =
-        'Admin, Dataverse; Owner, Dataverse, 2024, "Dataset created using the createDataset use case'
```
ekraffmiller, line 296: Can we make the expected year dynamic, so it won't need to be updated next year?

test/integration/collections/CollectionsRepository.test.ts
```diff
@@ -378,6 +378,8 @@ describe('CollectionsRepository', () => {
       expect(actualFilePreview.url).not.toBeUndefined()
       expect(actualFilePreview.releaseOrCreateDate).not.toBeUndefined()
       expect(actualFilePreview.type).toBe(CollectionItemType.FILE)
+      expect(actualFilePreview.restricted).toBe(false)
+      expect(actualFilePreview.canDownloadFile).toBe(true)
```
ekraffmiller, line 386: It would be nice to test the other updated values (variables, observations, tags and categories). If you upload a tab file, the variables and observations will be available. There is a test file, test-file-4.tab, available in the test-helpers/files directory that you can use.

ekraffmiller (commented):
Looks good! I just had some suggestions for the CollectionRespository tests

test/integration/collections/CollectionsRepository.test.ts
ChengShi-1: Hi ellen, thanks for reviewing! I tried to use `test-file-4.tab` to the test, and uploaded by `await uploadFileViaApi(testDatasetIds.numericId, 'test-file-4.tab', {categories: 'test', tabIngest: false})`, the third variable inside is for file metadata according to [native api](https://guides.dataverse.org/en/6.4/api/native-api.html#add-a-file-to-a-dataset), but the actualFilePreview still didn't show tabular tags, variables and observations. I checked the search api endpoint, the tabular data info shows correctly. I am not sure if there is something else I missed.

ekraffmiller: Hi Cheng, I think the problem may be that you are including `tabIngest: false`. The ingest process is needed to generate the variables and observations. Can you try it without that?

ChengShi-1: Thanks! but I tried delete the tabIngest, same result. I found that tabular info was only shown when its parent dataset published, I would try publish dataset first and see if the tests could work.

ekraffmiller (approved):
looks good, approved!

ofahimIQSS:
Visually inspected code, tests are passing. Merging PR

PR #238, opened 2025-01-16
Typescript strict config

### LINKED ISSUE #236: TS Config: strictNullChecks
Right now, the typescript configuration is not in strict mode and several important checks are not happening as they do in the frontend repo.

For example properties of an interface that are declared as `string | undefined` or `string | null` for example, are both treated as `string` and this exposes a risk when developing.

The goal of this issue is to configure typescript in a stricter mode and accommodate the code everywhere where after this change, typescript starts complaining.

g-saracca:
Applies strict typescript configuration to the project as we do in the spa frontend repo.
Now properties of an interface that are declared as `string | undefined` or `string | null` or `property?: string` for example, are treated as both values, so we can write more safe code now.
This PR also accommodate the code everywhere where after this config change, typescript was complaining because we weren't taking care of possibly undefined or null values, some values are casted with the `as` keyword because there where previous checks and it was safe to cast them.

- Closes #236 

Visual code inspection and validate that Github Action checks are OK.

### DIFF
--- a/src/collections/infra/repositories/CollectionsRepository.ts
+++ b/src/collections/infra/repositories/CollectionsRepository.ts
@@ -33,9 +33,9 @@ export interface NewCollectionContactRequestPayload {
 }
 
 export interface NewCollectionMetadataBlocksRequestPayload {
-  metadataBlockNames: string[]
-  facetIds: string[]
-  inputLevels: NewCollectionInputLevelRequestPayload[]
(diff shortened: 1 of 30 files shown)


ChengShi-1 (approved):
Good! Approve

ofahimIQSS:
Tests are passing - merging PR

PR #248, opened 2025-01-30
Get Dataset: add custom terms of use 

### LINKED ISSUE #247: Get Dataset Extension: add Custom Terms and convert to Markdown
If a Dataset has Custom Terms of Use, there are additional set of fields that associated with the custom terms, which are displayed in the Terms tab. 
They are: 

- termsOfUse
- confidentialDeclaration
- specialPermissions
- restrictions
- citationRequirements
- depositorRequirements
- conditions
- disclaimer

All these fields are textboxes, so should be converted to markdown if keepRawFields = false

ekraffmiller:
Needed for the Dataset Terms Tab, if custom terms are chosen, these additional fields are displayed.

- Closes #247 

Review the code and tests for Get Dataset use case


src/datasets/infra/repositories/transformers/DatasetPayload.ts
```diff
@@ -24,6 +24,14 @@ export interface DatasetPayload {
...
   sizeOfCollection?: string
   studyCompletion?: string
+  termsOfUse?: string
+  confidentialityDeclaration?: string
+  specialPermissions?: string
+  restrictions?: string
+  citationRequirements?: string
+  depositorRequirements?: string
+  conditions?: string
+  disclaimer?: string
```
g-saracca, line 34: Question, all these new properties could be undefined? or are we following some convention to treat these properties always as a string or undefined?

test/environment/.env
```diff
@@ -4,3 +4,4 @@ SOLR_VERSION=9.3.0
 DATAVERSE_IMAGE_REGISTRY=docker.io
 DATAVERSE_IMAGE_TAG=unstable
 DATAVERSE_BOOTSTRAP_TIMEOUT=5m
+STORAGE_DIR=/tmp/dataverse
```
g-saracca: what is this for?

g-saracca (commented):
Heey, just some minor comments.
To understand this, we need to format all these terms of use data into markdown to show it in a better way in the Dataset page?
I understand we wont need any change in the frontend repo as the default value of `keepRawFields` is `false`.

ekraffmiller:
> Heey, just some minor comments. To understand this, we need to format all these terms of use data into markdown to show it in a better way in the Dataset page? I understand we wont need any change in the frontend repo as the default value of `keepRawFields` is `false`.

Good morning, welcome back! Yes, all these fields are text boxes which allow html in the edit form, so they need to be displayed as markdown in the terms tab.  In my frontend PR for the terms tab, I'm using the MarkdownComponent to render them.

test/environment/.env
ekraffmiller: This is something I need to add locally for my build of js-dataverse to work, I committed it by mistake.  I've been meaning to ask about it - am I the only one having this issue?

src/datasets/infra/repositories/transformers/DatasetPayload.ts
ekraffmiller: That's a good point, all the fields are optional, except for termsOfUse, that is required, but only if the user chooses "Custom Terms" in the License field.  When the data is returned from the API, there is no specific field that specifies custom terms, that is determined by checking the license field - if the license object is undefined, then the terms are "custom", and the termsOfUse field is required.  Maybe there is a better way of enforcing that logic in the API?

test/environment/.env
g-saracca: I can confirm I don't have this issue on my end.

src/datasets/infra/repositories/transformers/DatasetPayload.ts
g-saracca: oh so its ok, all properties are optional. I dont' know about changing things in the API, just wanted to really know that these properties could really be undefined

g-saracca (approved):

ofahimIQSS:
Tests are passing - merging PR

PR #266, opened 2025-02-19
Restrict File use case

### LINKED ISSUE #259: Implement use case for restricting or unrestricting a single file
For SPA Q1 feature, "File Page: Edit File Options", we need a use case for restricting or unrestricting a single file.  The use case can use this API endpoint: https://guides.dataverse.org/en/latest/api/native-api.html#restrict-files.

g-saracca:
Adds the Restrict File use case. It can restrict or unrestrict a file.

- Closes #259 

Review code and tests.

### DIFF
--- a/src/core/infra/repositories/ApiRepository.ts
+++ b/src/core/infra/repositories/ApiRepository.ts
@@ -77,10 +77,8 @@ export abstract class ApiRepository {
     let requestData = data
 
     if (contentType === ApiConstants.CONTENT_TYPE_APPLICATION_JSON) {
-      if (typeof data === 'object') {
+      if (typeof data === 'object' || typeof data === 'boolean') {
         requestData = JSON.stringify(data)
-      } else if (typeof data === 'boolean') {
-        requestData = data.toString()
       }
     }
 
--- a/src/core/infra/repositories/ApiRepository.ts
+++ b/src/core/infra/repositories/ApiRepository.ts
@@ -20,7 +20,7 @@ export abstract class ApiRepository {
 
   public async doPost(
     apiEndpoint: string,
-    data: string | object,
+    data: string | object | boolean,
     queryParams: object = {},
     contentType: string = ApiConstants.CONTENT_TYPE_APPLICATION_JSON
   ): Promise<AxiosResponse> {
@@ -29,7 +29,7 @@ export abstract class ApiRepository {
 
   public async doPut(
     apiEndpoint: string,
-    data: string | object,
+    data: string | object | boolean,
     queryParams: object = {},
     contentType: string = ApiConstants.CONTENT_TYPE_APPLICATION_JSON
   ): Promise<AxiosResponse> {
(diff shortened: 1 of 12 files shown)


test/integration/files/FilesRepository.test.ts
```diff
@@ -725,4 +725,139 @@ describe('FilesRepository', () => {
...
+      try {
+        restrictFileDatasetIds = await createDataset.execute(TestConstants.TEST_NEW_DATASET_DTO)
+      } catch (error) {
+        throw new Error('Tests beforeEach(): Error while creating test dataset')
+      }
+      await uploadFileViaApi(restrictFileDatasetIds.numericId, testTextFile1Name).catch(() => {
+        throw new Error(`Tests beforeEach(): Error while uploading file ${testTextFile1Name}`)
+      })
+
+      await publishDatasetViaApi(restrictFileDatasetIds.numericId).catch(() => {
```
ekraffmiller: Is there a reason why the Dataset is published in this test?  The restrict action can be used on a Dataset when it is a draft version. Publishing also introduces wait times for the lock to be released, maybe this is contributing to the timing error I had in my local env?

src/files/domain/useCases/RestrictFile.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  /**
+   * Restrict or unrestrict an existing file.
+   * More detailed information about the file restriction behavior can be found in https://guides.dataverse.org/en/latest/api/native-api.html#restrict-files
+   *
+   * @param {number | string} [fileId] - The File identifier, which can be a string (for persistent identifiers), or a number (for numeric identifiers).
+   * @param {boolean} [restrict] - A boolean value that indicates whether the file should be restricted or unrestricted.
+   * @returns {Promise<void>} -This method does not return anything upon successful completion.
+   */
+  async execute(fileId: number | string, restrict: boolean): Promise<void> {
```
ekraffmiller, line 15: In the UI, the restrict file modal includes an option of changing the default fileAccessRequest value, and adding termsOfAccessForRestrictedFiles.  (these values are also in the Terms tab).  Do we want to add this as optional params to the restrict method?  Or from the frontend, should it be handled by making a separate API call to update Dataset Terms of Use?

Screenshot 2025-02-21 at 1 31 37 PM

src/core/infra/repositories/ApiRepository.ts
```diff
@@ -70,12 +70,20 @@ export abstract class ApiRepository {
...
     method: 'post' | 'put',
     apiEndpoint: string,
-    data: string | object,
+    data: string | object | boolean,
     queryParams: object = {},
     contentType: string = ApiConstants.CONTENT_TYPE_APPLICATION_JSON
   ): Promise<AxiosResponse> {
-    const requestData =
-      contentType == ApiConstants.CONTENT_TYPE_APPLICATION_JSON ? JSON.stringify(data) : data
+    let requestData = data
```
ekraffmiller, line 77: Is  this extra logic is needed? If `data === 'boolean'`, `JSON.stringify(data)` will  return the same thing as `data.toString()`

ekraffmiller (commented):
Looks good! Just a couple of questions

src/files/domain/useCases/RestrictFile.ts
g-saracca: Good question, the API is not accepting params and I think it's ok.
A second API call to update Dataset Terms of Use seems reasonable.
We can do that separately while working on the frontend integration. Is there a use case for doing that in Js-dataverse?

src/core/infra/repositories/ApiRepository.ts
g-saracca: You are right! thanks!

test/integration/files/FilesRepository.test.ts
g-saracca: You are right, I've removed the publishing!

g-saracca:
@ekraffmiller ready for review again.

src/files/domain/useCases/RestrictFile.ts
ekraffmiller: ok, we don't have a use case for updating terms of use yet, so yes would be good to work on that separately.

ekraffmiller (approved):
Looks good, approved!

ofahimIQSS:
tests are passing - merging pr

PR #273, opened 2025-03-04
DatasetValidator allow empty fields for conditionally required fields.

### LINKED ISSUE #272: DatasetValidator allow empty fields for conditionally required fields.
Modify the Create/Update Dataset Validator to allow empty values in child fields that are required but their parent fields (compound fields) are not.
These child fields should only be truly required when one of their siblings fields has a value; otherwise, they can be left empty.

g-saracca:
Modifies the Create/Update Dataset Validator to allow empty values in child fields that are required but their parent fields (compound fields) are not.
These child fields should only be truly required when one of their siblings fields has a value; otherwise, they can be left empty.

I have also added a new parameter to the `UpdateDataset` use case which is the `internalVersionNumber` of the dataset, and also now in the `GetDataset` use case we are returning that property to be used later. 
Integration tests were added 👍🏼 .

> * @param {number} [internalVersionNumber] - The internal version number of the dataset. If another user updates the dataset version metadata before you send the update request, data inconsistencies may occur. To prevent this, you can use the optional internalVersionNumber parameter. This parameter must include the internal version number corresponding to the dataset version being updated. Note that internal version numbers increase sequentially with each version update.

- Closes #272 

## Related Dataverse PRs:

- Depends on https://github.com/IQSS/dataverse/pull/11273

review code and unit tests


src/datasets/domain/models/Dataset.ts
```diff
@@ -5,6 +5,7 @@ export interface Dataset {
   persistentId: string
   versionId: number
   versionInfo: DatasetVersionInfo
+  internalVersionNumber: number
```
ChengShi-1, line 8: is this optional?  like` internalVersionNumber?: number, `since the internalVersionNumber is optional from api

src/datasets/domain/useCases/validators/SingleMetadataFieldValidator.ts
```diff
@@ -144,8 +150,58 @@ export class SingleMetadataFieldValidator extends BaseMetadataFieldValidator {
...
+        metadataFieldPosition: datasetMetadataFieldAndValueInfo.metadataFieldPosition,
+        allowEmptyForConditionallyRequiredField
       })
     }
   }
+
+  /**
+   * This method allows setting empty values for conditionally required child fields.
+   * A child field is conditionally required if it is required and its parent field is not required.
+   * The childfield should be required only if any of its sibling fields has a value, otherwise it should be optional.
```
ChengShi-1: childfield => child field?

src/datasets/infra/repositories/transformers/DatasetPayload.ts
```diff
@@ -5,6 +5,7 @@ export interface DatasetPayload {
   datasetId: number
   datasetPersistentId: string
   id: number
+  internalVersionNumber: number
```
ChengShi-1, line 8: is this optional? like `internalVersionNumber?: number`, since the internalVersionNumber is optional from api

src/datasets/domain/useCases/validators/SingleMetadataFieldValidator.ts
```diff
@@ -144,8 +150,58 @@ export class SingleMetadataFieldValidator extends BaseMetadataFieldValidator {
...
+            childMetadataFieldKey
+          ].isRequired
+      )
+    const hasConditionallyRequiredChildFields = Boolean(conditionallyRequiredChildFields)
+
+    if (
+      hasConditionallyRequiredChildFields &&
+      Object.values(conditionallyRequiredChildFields as string[]).includes(childMetadataFieldKey)
+    ) {
+      // At this point we know we are standing on a child field that is required and the parent field is not required
```
ChengShi-1, line 190: Not sure if you intentionally leave comments here?

test/integration/datasets/DatasetsRepository.test.ts
```diff
@@ -874,6 +913,104 @@ describe('DatasetsRepository', () => {
...
+        createdDataset.numericId,
+        DatasetNotNumberedVersion.LATEST,
+        false,
+        false
+      )
+      const actualCreatedDatasetInternalVersionNumber = actualCreatedDataset.internalVersionNumber
+
+      expect(actualCreatedDataset.internalVersionNumber).toBe(1)
+
+      // Now update the dataset and then update again with the same internal version number
```
ChengShi-1, line 979: There three comments here, I'm not sure if you intentionally leave comments here? cause they're actually helpful

ChengShi-1 (commented):
Good job! Just 2 small things.
1. I notice the internalVersion is an optional parameter from api, but it is set to be required in modal dataset.ts and datasetPayload.
2. Some comments left here, not sure if you want them to be a guidance, or you want to remove them. If it is meant to be there, just ignore my comments there.

src/datasets/domain/models/Dataset.ts
g-saracca: hi 👋🏼 , I think you are confusing the new optional parameter in the Update Dataset use case and this new property added to the Dataset model.

> I have also added a new parameter to the UpdateDataset use case which is the internalVersionNumber of the dataset, and also now in the GetDataset use case we are returning that property to be used later.
Integration tests were added 👍🏼 .

src/datasets/infra/repositories/transformers/DatasetPayload.ts
g-saracca: No, its not optional or it won't be undefined, this is the property returned from the API when getting a dataset.

src/datasets/domain/useCases/validators/SingleMetadataFieldValidator.ts
g-saracca: Yeah, intentionally for better understanding for someone else later.

test/integration/datasets/DatasetsRepository.test.ts
g-saracca: yep also intentionally.

src/datasets/domain/useCases/validators/SingleMetadataFieldValidator.ts
g-saracca: added space, child field

ChengShi-1 (approved):
Great

ofahimIQSS:
Tests are passing, merging

PR #285, opened 2025-04-11
Implement use case for editing tabular tags and categories

### LINKED ISSUE #257: Implement use case for editing categories and tabular tags
Inspired from [2025 Q1 proposal](https://docs.google.com/document/d/17U4evH3tZ1FZ_plwBDG2S1QpD4eTsWui6fm_XvXKLJ4/edit?tab=t.0#heading=h.urfyde7ncfu). The File  categories and tabular tags edit option for single file could be implemented since the api endpoints are ready [here.](https://guides.dataverse.org/en/6.4/api/native-api.html#updating-file-metadata)
depends on https://github.com/IQSS/dataverse/pull/11359
- edit categories
- edit tags if it's tabular

ChengShi-1:
For a single file
- edit categories
- edit tags if it's tabular

- Closes #https://github.com/IQSS/dataverse-client-javascript/issues/257

## Related Dataverse PRs:

- Depends on #

### DIFF
--- a/docs/useCases.md
+++ b/docs/useCases.md
@@ -1336,8 +1336,8 @@ const updateFileMetadataDTO = {
   restrict: false
 }
 
-await updateFileMetadata.execute(fileId, updateFileMetadataDTO).then((fileId) => {
-  console.log(`File updated successfully with file ID: ${fileId}`)
+await updateFileMetadata.execute(fileId, updateFileMetadataDTO).then(() => {
+  console.log(`File updated successfully`)
 })
 ```
 
@@ -1360,8 +1360,8 @@ const fileId: number | string = 123
 const categories = ['category 1', 'category 1']
 const replace = true
 
-await updateFileCategories.execute(fileId, categories, replace).then((fileId) => {
-  console.log(`File updated successfully with file ID: ${fileId}`)
+await updateFileCategories.execute(fileId, categories, replace).then(() => {
+  console.log(`File updated successfully`)
 })
 ```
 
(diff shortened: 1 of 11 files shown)


docs/useCases.md
```diff
@@ -1345,6 +1345,53 @@ _See [use case](../src/files/domain/useCases/UpdateFileMetadata.ts) implementati
...
+import { updateFileCategories } from '@iqss/dataverse-client-javascript'
+
+/* ... */
+
+const fileId: number | string = 123
+const categories = ['category 1', 'category 1']
+const replace = true
+
+await updateFileCategories.execute(fileId, categories, replace).then((fileId) => {
+  console.log(`File updated successfully with file ID: ${fileId}`)
```
g-saracca: In this example you are saying the execute returns a the `fileId` but in the use cases it shows like it returns `Promise<void>`, which one is it? by the tests I assume is the latter one?

test/functional/files/UpdateFileTabularTags.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    try {
+      await deleteCollectionViaApi(testCollectionAlias)
+    } catch (error) {
+      throw new Error('Tests afterAll(): Error while deleting test collection')
+    }
+  })
+
+  test('should successfully update tabular tags of a file', async () => {
```
g-saracca, line 62: For this  use case and also for UpdateFileCategories, could you write a test that asserts that when using replace in false or undefined, the tags or categories are added and they don't replace existing ones? e.g.: Dataset had [a,b,c] categories and you send [d,e], now categories are [a,b,c,d,e].
Also what would hapen if dataset had ['a', 'b'] and you send ['b', 'c'], what will happen wih duplicated 'b'? I think we could add lot more test cases here so we are more confident for the integration in the SPA later 👍🏼

g-saracca (changes requested):

ChengShi-1:
@g-saracca Hi German, changes made, please check again. Thanks for the review!

g-saracca (approved):

ofahimIQSS:
tests are passing, merging

PR #286, opened 2025-04-11
Implement use case to Delete dataset

### LINKED ISSUE #284: Implement use case for deleting a draft dataset
According the proposal and metrics of Q2, delete Draft Dataset is a must-have feature. 

When the dataset is published or not published yet, once it has a draft version, we should have a "Delete Dataset" function for the dataset. API has existed, so a use case for it is needed

ChengShi-1:
allow to delete a dataset 

[Native api reference](https://guides.dataverse.org/en/6.4/api/native-api.html#id78)

- Closes https://github.com/IQSS/dataverse-client-javascript/issues/284

## Related Dataverse PRs:

- Depends on #

### DIFF
--- a/docs/useCases.md
+++ b/docs/useCases.md
@@ -39,6 +39,7 @@ The different use cases currently available in the package are classified below,
     - [Update a Dataset](#update-a-dataset)
     - [Publish a Dataset](#publish-a-dataset)
     - [Deaccession a Dataset](#deaccession-a-dataset)
+    - [Delete a Dataset](#delete-a-dataset)
 - [Files](#Files)
   - [Files read use cases](#files-read-use-cases)
     - [Get a File](#get-a-file)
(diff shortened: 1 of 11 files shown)


ekraffmiller (commented):
Looks good!  The only general comment I have, is I think it would be good to change the name of the use case from `deleteDataset` to `deleteDatasetDraft`, to make it more clear that it is only the draft version of the dataset that will be deleted.  And in `useCases.md`, to add that an error will be thrown if a draft version of the dataset doesn't exist

ChengShi-1:
@ekraffmiller Hi ellen, thanks for reviewing. I changed the naming, and it's ready for review again

ekraffmiller (approved):
looks good, approved!

ofahimIQSS:
tests passing, merging.

PR #294, opened 2025-04-24
Implement GetMyDataCollectionItems use case

### LINKED ISSUE #283: Implement use case for getting My Data section data
Based on the My Data API, https://guides.dataverse.org/en/latest/api/native-api.html#mydata

This is needed for the Q2 deliverable of implementing the AccountPage: My Data section.

As part of this issue, we will analyze the requirements of the My Data section, and if needed request changes to the My Data API

ekraffmiller:
Implements a use case for calling the MyData API endpoint, used in the SPA Accounts Page -> My Data section.
Returns a collection of CollectionItems, similar to the GetCollectionItems use case.

- Closes #283 

Review code and tests


g-saracca:
Hi @ekraffmiller, should we already create a js-dataverse use case for this endpoint?
Doesn't it still have a lot of bugs to solve and things that might change because of them?
IMO it may be too early to try to make a Use Case and a frontend integration.

ekraffmiller:
> Hi @ekraffmiller, should we already create a js-dataverse use case for this endpoint? Doesn't it still have a lot of bugs to solve and things that might change because of them? IMO it may be too early to try to make a Use Case and a frontend integration.

That's true there are bug fixes to wait for, I just wasn't sure if we needed to work with what we have since I'm not sure when they will be fixed.  But I can put a waiting tag on this and work on other things until we decide what to do.

g-saracca:
> That's true there are bug fixes to wait for, I just wasn't sure if we needed to work with what we have since I'm not sure when they will be fixed. But I can put a waiting tag on this and work on other things until we decide what to do.

Ok, so if you did any temporary patch/fix to avoid the bugs please mention it in the PR. If you believe that there are no big patches to do to make it work it's ok otherwise I would wait but prioritising the API bug fixes to be fixed as soon as possible.

ekraffmiller:
> > That's true there are bug fixes to wait for, I just wasn't sure if we needed to work with what we have since I'm not sure when they will be fixed. But I can put a waiting tag on this and work on other things until we decide what to do.
> 
> Ok, so if you did any temporary patch/fix to avoid the bugs please mention it in the PR. If you believe that there are no big patches to do to make it work it's ok otherwise I would wait but prioritising the API bug fixes to be fixed as soon as possible.

Ok, to summarize the issues on the API - 

1. The per-page param doesn't work, it always returns 10 items.  [#10528](https://github.com/IQSS/dataverse/issues/10528)
2. The empty results are thrown as an error from the API.  [#11447](https://github.com/IQSS/dataverse/issues/11447)
3. There is no permissions check for using the otherIdentifier param.  [#11451](https://github.com/IQSS/dataverse/issues/11451)
4. The datasets in harvested dataverses are not being returned. [#11083](https://github.com/IQSS/dataverse/issues/11083)

These errors don't affect the logic of the use case, just the testing. When 2 and 3 are fixed, we need to update the tests.  Number 2 affects the SPA handling empty results.  So I think we can set it to waiting until 2 is fixed, as you said, it shouldn't be a big change.

g-saracca:
@ekraffmiller ok, it doesn't sound too bad right?
So I believe we can continue with this Use Case 👍🏼, and it should be a temporary workaround in the SPA integration regarding the 2nd item.
Thanks! I will review this PR

docs/useCases.md
```diff
@@ -215,6 +215,65 @@ This use case supports the following optional parameters depending on the search
 - **offset**: (number) Offset for pagination.
 - **collectionSearchCriteria**: ([CollectionSearchCriteria](../src/collections/domain/models/CollectionSearchCriteria.ts)) Supports filtering the collection items by different properties.
 
+#### List My Data Collection Items
```
g-saracca, line 219: Please add this header link to the Table of Contents.

src/collections/domain/models/CollectionItemSubset.ts
```diff
@@ -8,20 +8,26 @@ export interface CollectionItemSubset {
   totalItemCount: number
   countPerObjectType: CountPerObjectType
 }
+export interface MyDataCollectionItemSubset {
+  items: (CollectionPreview | DatasetPreview | FilePreview)[]
+  publishingFacet: CollectionItemsFacetLabel[]
```
g-saracca: I suppose the MyData endpoint doesn't send a `friendlyName` right? the `name` from the `CollectionItemsFacetLabel` will be friendly enough to display?

src/collections/infra/repositories/transformers/collectionTransformers.ts
```diff
@@ -140,3 +209,15 @@ const transformContactsPayloadToContacts = (
     displayOrder: contactPayload.displayOrder
   }))
 }
+
+const transformPublicationStatusResponseToLabels = (
```
g-saracca: You could just return the new Array of objects instead of pushing:
```
const transformPublicationStatusResponseToLabels = (
  publicationStatusCounts: MyDataPublicationStatusCountsPayload
): CollectionItemsFacetLabel[] => [
  { name: 'Published', count: publicationStatusCounts.published_count },
  { name: 'Unpublished', count: publicationStatusCounts.unpublished_count },
  { name: 'Draft', count: publicationStatusCounts.draft_count },
  { name: 'In Review', count: publicationStatusCounts.in_review_count },
  { name: 'Deaccessioned', count: publicationStatusCounts.deaccessioned_count },
]
```

test/functional/datasets/PublishDataset.test.ts
```diff
@@ -67,7 +67,7 @@ describe('execute', () => {
       createdDatasetIdentifiers.persistentId,
       VersionUpdateType.MAJOR
     )
-    await waitForNoLocks(createdDatasetIdentifiers.numericId, 10)
+    await waitForNoLocks(createdDatasetIdentifiers.numericId, 30)
```
g-saracca: Just curious on why this specific test needs more waiting time?

g-saracca (changes requested):
Nice job, looks great! 🎉  Just some minor comments.
It's awesome that MyData endpoint returns the same properties(different naming) as the Search endpoint so we can share Collection, Dataset and File Previews models between these two use cases!
One more thing, could you add integration tests regarding filtering in the `GetMyDataCollectionItems` use case as we have with the `GetCollectionItems` use case?

src/collections/domain/models/CollectionItemSubset.ts
ekraffmiller: After working on the frontend for this, I'm thinking maybe it would be better to return a generic list of facets, rather than specifically the publishing facet, that way we can reuse the CollectionItemSubset for both GetCollectionItems and GetMyDataCollectionItems.  
What do you think? We would lose the specificity, but it also allows flexibility if we return other facets in the future.

g-saracca: Yes, it's a good idea, it could be good to share the same model between these two although it comes with a big responsibility, we don't want to end up with too many optional properties just because they may come in one endpoint and not in the other.
With the info the myData returns about the facets, are you going to be able to map it to the same model as facets in `GetCollectionItems`?

ekraffmiller: Yes, that was my idea to map it to the same model as the facets in `GetCollectionItems`

ekraffmiller:
Hi @g-saracca I finally fixed the tests!  when I used createCollectionViaApi, the collection didn't appear in the results, but when I used the use case instead (createCollection.execute()) it works. I don't know why, since the collection seems to be created correctly in both instances.  Also, when I create the new user, I have to give it superuser privileges in order to have the permission to create a subcollection within the test collection.

test/testHelpers/users/apiTokenHelper.ts
```diff
@@ -17,11 +20,16 @@ export const createApiTokenViaApi = async (userName: string): Promise<string> =>
...
+    const token = await axios
       .get(`${TestConstants.TEST_API_URL}/builtin-users/${userName}/api-token?password=${userName}`)
       .then((response) => response.data.data.message)
-  } catch (error) {
-    console.log(error)
+    if (createSuperUser) {
+      await axios.put(`${TestConstants.TEST_API_URL}/admin/superuser/${userName}`, 'true')
+    }
+    return token
+  } catch (error: Error | any) {
```
g-saracca: lint warning here, try with `(error: Error | unknown)` and if you want to keep the log change it to `console.log(error?.message)`

ekraffmiller: When I try adding `console.log(error?.message)` I get another error:
`TS2339: Property message does not exist on type {}`.
I just added this for debugging test so I think it's ok to remove it

g-saracca: yep, ok so I think it's ok to just remove

g-saracca (approved):
Approving!

ofahimIQSS:
Tests are passing - merging.

PR #298, opened 2025-05-05
Fix the integration test for DirectUpload.test.ts 

ekraffmiller:
Remove the test of the primary key value, since we can't guarantee the generated value


g-saracca (approved):
LGTM

PR #302, opened 2025-05-09
Use Case File Has Been Deleted

### LINKED ISSUE #301: File Has Been Deleted Use Case
Create a new use case based on this [API endpoint](https://guides.dataverse.org/en/latest/api/native-api.html#file-has-been-deleted) to know if a particular file that existed in a previous version of the dataset no longer exists in the latest version.
This is needed for the SPA to know when to allow the File edition in both file page or file in a datset page.

ChengShi-1:

Create a new use case based on this [API endpoint](https://guides.dataverse.org/en/latest/api/native-api.html#file-has-been-deleted) to know if a particular file that existed in a previous version of the dataset no longer exists in the latest version.
This is needed for the SPA to know when to allow the File edition in both file page or file in a datset page.

- Closes #301 

## Related Dataverse PRs:

- Depends on #

### DIFF
--- a/docs/useCases.md
+++ b/docs/useCases.md
@@ -52,6 +52,7 @@ The different use cases currently available in the package are classified below,
     - [Get the size of Downloading all the files of a Dataset Version](#get-the-size-of-downloading-all-the-files-of-a-dataset-version)
     - [Get User Permissions on a File](#get-user-permissions-on-a-file)
     - [List Files in a Dataset](#list-files-in-a-dataset)
+    - [Is File Deleted](#is-file-deleted)
   - [Files write use cases](#files-write-use-cases)
     - [File Uploading Use Cases](#file-uploading-use-cases)
     - [Delete a File](#delete-a-file)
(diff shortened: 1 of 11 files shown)


docs/useCases.md
```diff
@@ -1571,6 +1571,30 @@ If restrict is false then enableAccessRequest and termsOfAccess are ignored
...
+##### Example call:
+
+```typescript
+import { fileHasBeenDeleted } from '@iqss/dataverse-client-javascript'
+
+/* ... */
+
+const fileId = 12345
+
+await fileHasBeenDeleted.execute(fileId).then((hasBeenDeleted: boolean) => {
```
g-saracca: should we name this use case that returns a boolean like this or `getFileHasBeenDeleted`? 🤔
@GPortas & @ekraffmiller  what do you think?
This is I believe the first use case that returns a boolean so it could serve as the basis for future use cases.

test/integration/files/FilesRepository.test.ts
```diff
@@ -840,6 +840,81 @@ describe('FilesRepository', () => {
...
+    })
+
+    test('should return error if the dataset is unpublished and the file has been deleted', async () => {
+      await sut.deleteFile(fileId)
+
+      const expectedError = new ReadError(`[404] File with ID ${nonExistentFiledId} not found.`)
+      await expect(sut.fileHasBeenDeleted(nonExistentFiledId)).rejects.toThrow(expectedError)
+    })
+
+    test('should return True when the dataset is published and the file has not been deleted', async () => {
```
g-saracca: could we change the naming of this test, is a little bit confusing, you are testing both first that is not deleted and then that is indeed deleted.

g-saracca (changes requested):
@ChengShi-1 just two minor comments, one above about a naming convention and could you please add also an integration test to check whether this use case / api endpoint returns true or false when a file has been replaced.
I believe it should return true also when a file has been replaced and we need that behaviour for the SPA.

g-saracca (approved):
Looks great! thanks for applying the changes cheng 🙏🏼

ofahimIQSS:
nothing suspicious going on here - things look to be passing. Merged!

PR #306, opened 2025-05-23
GetCollectionItems use case DatasetPreview publicationDate

g-saracca:
Maps and send the publicationDate of a dataset in the GetCollectionItems DatasetPreview information.

To test: is just a new property, code review and check that tests are passing.


g-saracca:
Closing, this is not needed.

PR #307, opened 2025-05-27
Avoid double encoding of search term

g-saracca:
Avoids double encoding in the search api and my data api related use cases.
We were using `URLSearchParams` for query params and also `encodeURIComponent` for the search text so we were double encoding the `mydata_search_term` and the `q` query params.

Code review and tests passing.

No
## Additional documentation:
No


ChengShi-1 (approved):
looks good

ofahimIQSS:
Merging as tests are passing. Thanks for the fix.

PR #313, opened 2025-06-02
Change return type of getMyDataCollectionItems use case

### LINKED ISSUE #304: Update the return type of getMyData use case
The return type of getMyDataCollectionItems is CollectionItemSubset, which is shared with getCollectionItems.  This works by formatting the PublicationStatus results as facets, which was meant to be extensible, but is awkward to use in the frontend code and is a different format from the data returned from the dataverse API.  

So this issue it to update the return type so it returns a list of PublicationStatus counts, rather than facets.

ekraffmiller:
Update the return type of the use case so it more closely matches what is being returned by the API

- Closes #304 

Note that this is a breaking change, so it would be good to wait to merge this until the PR for the frontend change is ready also: https://github.com/IQSS/dataverse-frontend/issues/734.  

Review test code


src/core/domain/models/PublicationStatus.ts
```diff
@@ -1,7 +1,7 @@
 export enum PublicationStatus {
+  Deaccessioned = 'Deaccessioned',
+  InReview = 'In Review',
   Published = 'Published',
   Unpublished = 'Unpublished',
-  Draft = 'Draft',
-  Deaccessioned = 'Deaccessioned',
-  InReview = 'In Review'
+  Draft = 'Draft'
```
g-saracca: Can you keep same order as before? 
```
export enum PublicationStatus {
  Published = 'Published',
  Unpublished = 'Unpublished',
  Draft = 'Draft',
  Deaccessioned = 'Deaccessioned',
  InReview = 'In Review'
}
```

g-saracca (dismissed):

ekraffmiller:
Hi @g-saracca, I had to update the imports for PublicationStatus to get this to work in my frontend PR (https://github.com/IQSS/dataverse-frontend/pull/737).  It seems that including 'src' in the relative path doesn't cause a compile error, but causes an error when using the library.  Can you review again please?   (I also added PublicationStatus to the exports, but I'm not sure if I need to do that)

g-saracca (dismissed):
@ekraffmiller, looks good and yes, you only need to export a type from the library if you need it from the consumer application. In the [SPA side we have our own](https://github.com/IQSS/dataverse-frontend/blob/develop/src/shared/core/domain/models/PublicationStatus.ts) PublicationStatus enum, but anyways is ok if you leave that enum export

g-saracca (approved):

ofahimIQSS:
Tests passing- merging

PR #318, opened 2025-06-24
Add GetUserSelectableRoles use case

### LINKED ISSUE #300: Create User Roles use case
Create a use case that calls a new API endpoint that will return a list of roles that the user has been assigned in the dataverse. Depends on this Dataverse issue: https://github.com/IQSS/dataverse/issues/11434

ekraffmiller:
Created a use case for the new API end point, get user selectable roles (api/roles/userSelectable). 

- Closes #300

### DIFF
--- a/src/roles/domain/repositories/transformers/roleTransformers.ts
+++ b/src/roles/domain/repositories/transformers/roleTransformers.ts
@@ -1,10 +1,11 @@
 import { AxiosResponse } from 'axios'
 import { Role } from '../../models/Role'
+import { RolePayload } from './RolePayload'
 
 export const transformRolesUserSelectableResponseToRoles = (response: AxiosResponse): Role[] => {
-  const roleUserSelectablePayload = response.data.data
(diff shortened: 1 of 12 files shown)


src/roles/domain/repositories/transformers/roleTransformers.ts
```diff
@@ -0,0 +1,14 @@
+import { AxiosResponse } from 'axios'
+import { Role } from '../../models/Role'
+
+export const transformRolesUserSelectableResponseToRoles = (response: AxiosResponse): Role[] => {
+  const roleUserSelectablePayload = response.data.data
+
+  return roleUserSelectablePayload.map((role: any) => ({
+    id: role.id,
```
g-saracca, line 9: Hi @ekraffmiller , I only found this type warning and we are good to go.
I think we can avoid the `any` type here if you create a file here in the `transformers` folder with the type of the `RoleUserSelectablePayload`

g-saracca (approved):

ofahimIQSS:
Looks good, merging.

PR #340, opened 2025-07-22
Get Citation In Other Format Use Case

### LINKED ISSUE #341: Get Citation In Other Formats Use Case
Inspired by 2025 Q3 SPA proposal, we could add the citation in other formats use case. [Native API works](https://guides.dataverse.org/en/latest/api/native-api.html#:~:text=%26includeDeaccessioned%3Dtrue%22-,Get%20Citation%20In%20Other%20Formats,-Dataverse%20can%20also)

Image

ChengShi-1:
Inspired by 2025 Q3 SPA proposal, we could add the citation in other formats use case. [Native API works](https://guides.dataverse.org/en/latest/api/native-api.html#:~:text=%26includeDeaccessioned%3Dtrue%22-,Get%20Citation%20In%20Other%20Formats,-Dataverse%20can%20also)

The return results have different content types, include 'EndNote' (XML), 'RIS' (plain text), 'BibTeX' (plain text), 'CSLJson' (JSON), and 'Internal' (HTML). 
image

- Closes IQSS/dataverse-client-javascript#341 

## Related Dataverse PRs:

- Depends on #

The use case will be used for citation downloads(like the screenshot). The return type is still `String` but in different Content-Types, I think it makes more sense to have an use case to fetch the information here, and write a file to download by users in the front end repo?

RoleHelper for testing changed because[ Revert "feat: "Link Dataset/Dataverse" permission"](https://github.com/IQSS/dataverse/pull/11682/files#diff-f7f8c09756ce44634bf7d9a243e628166a608079fffe7cd71f77f56f949114f6)


src/datasets/domain/models/CitationFormats.ts
```diff
@@ -0,0 +1,7 @@
+export enum CitationFormats {
```
ekraffmiller:
```suggestion
export enum CitationFormat {
```
To be consistent with other enums, I think it's better to use a singular name, rather than plural

src/datasets/domain/models/CitationResponse.ts
```diff
@@ -0,0 +1,4 @@
+export type CitationResponse = {
+  content: string | object
```
ekraffmiller: Since we are doing a client-side download,  I think we will have to stringify the json object if type = application/json.  So rather than return either a string or object, it might be more straightforward to stringify the json object response from the API, and content then can always be type string.
```suggestion
  content: string
```

src/datasets/domain/models/CitationResponse.ts
```diff
@@ -0,0 +1,4 @@
+export type CitationResponse = {
```
ekraffmiller: Another naming suggestion... 'Response' doesn't really describe the object, so maybe instead the name could be something like 'FormattedCitation'?

ekraffmiller (commented):
Looks good! I just have suggestions for naming and the domain model.

ChengShi-1:
@ekraffmiller Thanks for suggestions, Ellen! just fixed names and stringify json object

ekraffmiller (dismissed):
looks good, approved!

ChengShi-1:
@ekraffmiller Hi Ellen, I add one commit for exporting this use case in `dataset/index.ts`, and allow persistent id. Could you check and approve it again? Thanks!

ekraffmiller (approved):
looks good, approved 👍

ofahimIQSS:
tests passing, merging

PR #360, opened 2025-09-08
specify datasetType when creating a dataset

### LINKED ISSUE #359: allow a dataset to be created with a dataset type other than "dataset" (e.g. "software", "workflow", "review", etc.)
Ever since https://github.com/IQSS/dataverse/pull/10694 was merged, an implicit dataset type of "dataset" has been persisted to the database when a dataset is created.

As explained at https://guides.dataverse.org/en/6.7.1/api/native-api.html#create-a-dataset-with-a-dataset-type-software-etc it's possible to specify an alternate dataset type such as "software", "workflow", "review", etc. if the Dataverse backed can been configured to support them (see https://guides.dataverse.org/en/6.7.1/api/native-api.html#list-dataset-types ).

The JSON in the guides (see above) starts like this:

```
{
  "datasetType": "software",
  "datasetVersion": {
    "license": {
      "name": "CC0 1.0",
      "uri": "http://creativecommons.org/publicdomain/zero/1.0"
    },
    "metadataBlocks": {
      "citation": {
        "fields": [
          {
            "value": "pyDataverse",
            "typeClass": "primitive",
            "multiple": false,
            "typeName": "title"
          },
...
```

The use case for the "review" dataset type is described in this issue about Trusted Data:

- https://github.com/IQSS/dataverse-pm/issues/425

pdurbin:

At dataset creation time, we need to allow `datasetType` to be passed so that types other than the default ("dataset") can be persisted to the database.

- Intended for this issue but we'll make a fresh PR: #359

## Related Dataverse PRs:

- Frontend PR that that uses this PR: https://github.com/IQSS/dataverse-frontend/pull/822
- Parent project, Trusted Data: https://github.com/IQSS/dataverse-pm/issues/425

We also need to support listing dataset types ( https://guides.dataverse.org/en/6.7.1/api/native-api.html#list-dataset-types ) but perhaps that should be a separate issue and pull request?

- On the backend, create a datasetType such as "software" or "review": https://guides.dataverse.org/en/6.7.1/api/native-api.html#add-dataset-type
- Use that type to create a dataset of that type by supplying it in the `datasetType` field. In the JSON, it looks like this:

```
{
  "datasetType": "software",
  "datasetVersion": {
    "license": {
      "name": "CC0 1.0",
      "uri": "http://creativecommons.org/publicdomain/zero/1.0"
    },
    "metadataBlocks": {
      "citation": {
        "fields": [
          {
            "value": "pyDataverse",
            "typeClass": "primitive",
            "multiple": false,
            "typeName": "title"
          },
...
```

Yes, we should add something like this:

You can now specify a datasetType when creating a dataset.

## Additional documentation:

None.

### DIFF
--- a/src/datasets/domain/useCases/CreateDataset.ts
+++ b/src/datasets/domain/useCases/CreateDataset.ts
@@ -20,6 +20,7 @@ export class CreateDataset extends DatasetWriteUseCase<CreatedDatasetIdentifiers
    *
    * @param {DatasetDTO} [newDataset] - DatasetDTO object including the new dataset metadata field values for each metadata block.
    * @param {string} [collectionId] - Specifies the collection identifier where the new dataset should be created (optional, defaults to :root).
+   * @param {string} [datasetType] - Specifies the dataset type (optional, when omitted, defaults to "dataset").
    * @returns {Promise<CreatedDatasetIdentifiers>}
    * @throws {ResourceValidationError} - If there are validation errors related to the provided information.
    * @throws {ReadError} - If there are errors while reading data.
--- a/src/datasets/domain/useCases/CreateDataset.ts
+++ b/src/datasets/domain/useCases/CreateDataset.ts
@@ -27,10 +27,16 @@ export class CreateDataset extends DatasetWriteUseCase<CreatedDatasetIdentifiers
    */
   async execute(
     newDataset: DatasetDTO,
-    collectionId = ROOT_COLLECTION_ID
+    collectionId = ROOT_COLLECTION_ID,
+    datasetType?: string
   ): Promise<CreatedDatasetIdentifiers> {
     const metadataBlocks = await this.getNewDatasetMetadataBlocks(newDataset)
     this.getNewDatasetValidator().validate(newDataset, metadataBlocks)
-    return this.getDatasetsRepository().createDataset(newDataset, metadataBlocks, collectionId)
+    return this.getDatasetsRepository().createDataset(
+      newDataset,
+      metadataBlocks,
+      collectionId,
+      datasetType
+    )
   }
 }
--- a/test/unit/datasets/CreateDataset.test.ts
+++ b/test/unit/datasets/CreateDataset.test.ts
@@ -51,7 +51,8 @@ describe('execute', () => {
     expect(datasetsRepositoryStub.createDataset).toHaveBeenCalledWith(
       testDataset,
       testMetadataBlocks,
-      ROOT_COLLECTION_ID
+      ROOT_COLLECTION_ID,
+      undefined
     )
   })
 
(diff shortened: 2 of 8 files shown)


src/datasets/domain/useCases/CreateDataset.ts
```diff
@@ -27,10 +27,16 @@ export class CreateDataset extends DatasetWriteUseCase<CreatedDatasetIdentifiers
    */
   async execute(
     newDataset: DatasetDTO,
-    collectionId = ROOT_COLLECTION_ID
+    collectionId = ROOT_COLLECTION_ID,
+    datasetType?: string
```
ChengShi-1, line 32: Could you add the param to the description above  @param {string}...

test/unit/datasets/CreateDataset.test.ts
```diff
@@ -51,7 +51,8 @@ describe('execute', () => {
     expect(datasetsRepositoryStub.createDataset).toHaveBeenCalledWith(
       testDataset,
       testMetadataBlocks,
-      ROOT_COLLECTION_ID
+      ROOT_COLLECTION_ID,
+      undefined
```
ChengShi-1, line 55: You could add an additional test to make sure datasetType would be called here. 
For example, 
you could create a  function `createDatasetwithDatasetType` in `DataseHelper` and do something similar with this
`
 expect(datasetsRepositoryStub.createDatasetwithDatasetType).toHaveBeenCalledWith(
      testDataset,
      testMetadataBlocks,
      ROOT_COLLECTION_ID,
      datasetType
    )`

ChengShi-1 (commented):
Looks awesome!  

Could you also add or edit test cases in integration test `test/integration/datasets/DatasetsRepository.test.ts` and functional test `test/functional/datasets/CreateDataset.test.ts`? it may be harder to test if we don't have datasetType on DatasetDTO and dataset model. 
Also, the `useCase.md` may need to be updated with new parameter.

src/datasets/domain/useCases/CreateDataset.ts
pdurbin: Sure, added in 156963d.

test/unit/datasets/CreateDataset.test.ts
pdurbin: Good idea. Added in 48cc5dc. Thanks for the help with this!

pdurbin:
Closing. As discussed with @ChengShi-1 we're thinking a better order for PRs would be:

- get dataset types: for this I created a new PR: #366
- add, modify and delete dataset types (I'll work on this next)
- save a dataset with a specific dataset type (this PR, but I'll make a new one after I'm done with the PRs above)

PR #385, opened 2025-10-02
Removed date field validations used in create and update dataset use cases

g-saracca:
Removed date field validations used in create and update dataset use cases, since validation is already handled in the frontend  and the backend (other clients should perform client side validation also if they want). This avoids duplicated logic and keeps the package focused on its core responsibility.

Not much to test, just removing a validation, check that tests keep passing.

Yes, added to the changelog

## Additional documentation:
N/A


ChengShi-1 (approved):

ofahimIQSS:
looks good, merging

PR #391, opened 2025-10-17
Use Case Creating Dataset Template

### LINKED ISSUE #390: Create a Collection Template Use Case
Related Dataverse PR: https://github.com/IQSS/dataverse/pull/11565
Parent Issue https://github.com/IQSS/dataverse/issues/11453 (some other endpoints are needed from backend)

POST /dataverses/{id}/templates: Creates a template for a given Dataverse collection id.

ChengShi-1:
 Creates a template for a given Dataverse collection id.

- Closes #390 

## Related Dataverse PRs: 

- Depends on #https://github.com/IQSS/dataverse/pull/11565

Get dataset template places in datasetsRepository, but I place this to collection repository. I am not sure if I should put thi in dataset or collection repository, and I am open to move it to dataset repository if you think it's better 

yes


docs/useCases.md
```diff
@@ -18,6 +18,7 @@ The different use cases currently available in the package are classified below,
     - [Get Collection Featured Items](#get-collection-featured-items)
     - [Get Collections for Linking](#get-collections-for-linking)
   - [Collections write use cases](#collections-write-use-cases)
+    - [Create a Template](#create-a-template)
```
g-saracca: Could you move this to the last position of the Collections write use cases?

src/collections/domain/useCases/CreateTemplate.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  constructor(collectionsRepository: ICollectionsRepository) {
+    this.collectionsRepository = collectionsRepository
+  }
+
+  /**
+   * Creates a Dataset Template in the specified collection.
+   *
+   * @param {TemplateCreateDTO} template - Template definition payload.
+   * @param {number | string} [collectionIdOrAlias = ':root'] - Target collection id or alias.
```
g-saracca:
```suggestion
   * @param {number | string} [collectionIdOrAlias = ':root'] - A generic collection identifier, which can be either a string (for queries by CollectionAlias), or a number (for queries by CollectionId)
   * If this parameter is not set, the default value is: ':root'
```

src/collections/domain/dtos/TemplateCreateDTO.ts
```diff
@@ -0,0 +1,45 @@
...
+  name: string
+  isDefault: boolean | false // defaults to false if not provided
+  fields?: TemplateFieldDTO[]
+  instructions?: TemplateInstructionDTO[]
+}
+
+export interface TemplateFieldDTO {
+  typeName: string
+  multiple: boolean
+  typeClass: MetadataFieldTypeClass
```
g-saracca: Do we really need to send the `multiple` and `typeClass` values to the endpoint for creating a template?
For the Get Dataset Templates use case I've used this [DTO](https://github.com/IQSS/dataverse-client-javascript/blob/406416c15b0ee5914a874c8dc06e166090642083/test/testHelpers/datasets/datasetTemplatesHelper.ts#L5) and didn't encounter any issue

src/collections/domain/dtos/TemplateCreateDTO.ts
```diff
@@ -0,0 +1,45 @@
+import { MetadataFieldTypeClass } from '../../../metadataBlocks/domain/models/MetadataBlock'
+
+export interface TemplateCreateDTO {
+  name: string
+  isDefault: boolean | false // defaults to false if not provided
```
g-saracca: why narrowing as `| false`?
Would not be more convenient to make it optional in that case? `isDefault?: boolean`
In that case not sending it will defaults to false in the backend logic right?

g-saracca (changes requested):
I’ve left some comments and questions, and I have one more about where to place the “create template” use case. You might have noticed that I put the “get dataset templates” in the Datasets repository, what do you think we should do with this one? If we keep it here under Collections, it might be worth refactoring later to move the get dataset templates into the Collections repository as well? We could ask about this in the dv-core-spa channel.

src/collections/domain/dtos/TemplateCreateDTO.ts
ChengShi-1: It's set to false because I mock the behavior from the backend. I'll change to optional, which make better sense

g-saracca (approved):
Looks good!

ofahimIQSS:
tests passing - merging

PR #392, opened 2025-10-20
feat: introduce DatasetTypeDTO and remove optional from id property in DatasetType model

g-saracca:
Adds a new DTO interface used to define the object structure requested for creating a dataset type.
With this change we don't use the `DatasetType` one for this same purpose and we just leave it for defining the structure of a dataset type after it was created.

Check tests are passing.

No.


pdurbin (approved):
No objection.

g-saracca:
Merging as it's a typescript types only change.

PR #395, opened 2025-10-23
Add pagination query parameters to Dataset&File Version Summaries use case 

### LINKED ISSUE #394: Add pagination query parameters to Dataset&File Version Summaries use case
Since we had an API Refactoring and enhancement from Dataverse API https://github.com/IQSS/dataverse/pull/11859, we need to update Dataset and File Version Summaries query parameter

- limit: The maximum number of version differences to return.
- offset: The number of version differences to skip from the beginning of the list. Used for retrieving subsequent pages of results.

For example, to get the second page of results, with 2 items per page, you would use limit=2 and offset=2 (skipping the first two results).

For datasets:
`curl -H "X-Dataverse-key: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X GET "https://demo.dataverse.org/api/datasets/:persistentId/versions/compareSummary?persistentId=doi:10.5072/FK2/BCCP9Z&limit=2&offset=2"`

For files:
`curl -H "X-Dataverse-key: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx" -X GET "https://demo.dataverse.org/api/files/1234/versionDifferences?limit=2&offset=2"`

ChengShi-1:
Add pagination query parameters to Dataset&File Version Summaries use case 

- Closes #394 

## Related Dataverse PRs:

- Depends on #https://github.com/IQSS/dataverse/pull/11859

yes

### DIFF
--- a/src/datasets/infra/repositories/DatasetsRepository.ts
+++ b/src/datasets/infra/repositories/DatasetsRepository.ts
@@ -310,10 +310,10 @@ export class DatasetsRepository extends ApiRepository implements IDatasetsReposi
     limit?: number,
     offset?: number
   ): Promise<DatasetVersionSummaryInfo[]> {
-    const queryParams: { per_page?: string; start?: string } = {}
-
-    limit && (queryParams.per_page = limit.toString())
-    offset && (queryParams.start = offset.toString())
+    const queryParams: { limit?: string; offset?: string } = {
+      limit: limit?.toString(),
+      offset: offset?.toString()
+    }
 
     return this.doGet(
       this.buildApiEndpoint(this.datasetsResourceName, 'versions/compareSummary', datasetId),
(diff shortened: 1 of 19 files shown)


src/datasets/infra/repositories/DatasetsRepository.ts
```diff
@@ -306,11 +306,19 @@ export class DatasetsRepository extends ApiRepository implements IDatasetsReposi
...
   public async getDatasetVersionsSummaries(
-    datasetId: string | number
+    datasetId: string | number,
+    limit?: number,
+    offset?: number
   ): Promise<DatasetVersionSummaryInfo[]> {
+    const queryParams: { limit?: string; offset?: string } = {
+      limit: limit?.toString(),
+      offset: offset?.toString()
+    }
```
g-saracca, line 321: You can follow same approach as [here](https://github.com/IQSS/dataverse-client-javascript/blob/3407247f00ed4f3ef811c0852cb274a7c648f050/src/collections/infra/repositories/CollectionsRepository.ts#L190) for limit and offset, using URLSearchParams

g-saracca (changes requested):
Just one comment about how to handle query params and one question, are we returning the total count from these use cases, similar how we do for the GetCollectionItems?
You will need the total count to know when to keep asking or when you reach the final page.

src/datasets/infra/repositories/DatasetsRepository.ts
ChengShi-1: Yes, the api returns totalCount but I didn't add it to use case. I modified the code to apply totalCount, please review again. Thank you! @g-saracca

g-saracca (approved):

g-saracca:
@ChengShi-1 approved, note that this is a breaking change, you will need a frontend change to avoid things breaking 👍🏼

ofahimIQSS:
awesome - merging!

PR #397, opened 2025-11-04
feat: add config setting "bearerTokenGetFunction"

### LINKED ISSUE #396: Feature request: custom config for retrieving bearer token
We'd like to have a bearer token configuration that is more flexible than the current `bearerTokenLocalStorageKey`, for example, by allowing a custom function to retrieve the bearer token.

Our app uses [@axa-fr/react-oidc](https://github.com/AxaFrance/oidc-client/tree/main/packages/react-oidc), which stores the access token within a JSON object in local storage, not a simple string. This is not supported by `bearerTokenLocalStorageKey`.

Image

vera:

This PR adds a config setting `bearerTokenGetFunction` to define a custom function to retrieve the bearer token. This is more flexible than the current `bearerTokenLocalStorageKey` setting.

- Closes #396

## Related Dataverse PRs:

/

/

I've tested this by initializing with the following config:

```
ApiConfig.init(
  "http://localhost:8080/api",
  DataverseApiAuthMechanism.BEARER_TOKEN,
  undefined,
  undefined,
  () => {
    const value = localStorage.getItem("oidc.default");
    if (value) {
      const token = JSON.parse(value).tokens?.accessToken;
      return token || null;
    } else {
      return null;
    }
  }
);
```

and confirmed that bearer-token-authenticated requests are correctly sent.

Yes, that would be good.

## Additional documentation:

/

### DIFF
--- a/src/core/infra/repositories/ApiConfig.ts
+++ b/src/core/infra/repositories/ApiConfig.ts
@@ -10,7 +10,7 @@ export class ApiConfig {
     dataverseApiAuthMechanism: DataverseApiAuthMechanism,
     dataverseApiKey?: string,
     bearerTokenLocalStorageKey?: string,
-    bearerTokenGetFunction?: () => string
+    bearerTokenGetFunction?: () => string | null
   ) {
     this.dataverseApiUrl = dataverseApiUrl
     this.dataverseApiAuthMechanism = dataverseApiAuthMechanism
--- a/src/core/infra/repositories/ApiConfig.ts
+++ b/src/core/infra/repositories/ApiConfig.ts
@@ -3,17 +3,20 @@ export class ApiConfig {
   static dataverseApiAuthMechanism: DataverseApiAuthMechanism
   static dataverseApiKey?: string
   static bearerTokenLocalStorageKey?: string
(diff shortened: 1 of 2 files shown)


src/core/infra/repositories/ApiConfig.ts
```diff
@@ -3,17 +3,20 @@ export class ApiConfig {
...
   static bearerTokenLocalStorageKey?: string
+  static bearerTokenGetFunction?: () => string | null
 
   static init(
     dataverseApiUrl: string,
     dataverseApiAuthMechanism: DataverseApiAuthMechanism,
     dataverseApiKey?: string,
-    bearerTokenLocalStorageKey?: string
+    bearerTokenLocalStorageKey?: string,
+    bearerTokenGetFunction?: () => string
```
ekraffmiller: I think the return type should match the property declaration
```suggestion
    bearerTokenGetFunction?: () => string | null
```

ekraffmiller (commented):
Hi Vera, this looks good, I just suggested a small change in the init() function.

vera:
Thanks, I accepted the suggestion!

ekraffmiller (approved):
looks good, approved!

PR #402, opened 2025-11-27
Add use case for getting dataset versions

vera:

This PR adds code for using the ["List Versions of a Dataset" API](https://guides.dataverse.org/en/latest/api/native-api.html#list-versions-of-a-dataset).

I'm not aware of an open issue.

## Related Dataverse PRs:

- ~Depends on https://github.com/IQSS/dataverse/pull/12008~

Since during review of https://github.com/IQSS/dataverse/pull/12008, it was decided not to add the `isInReviewState` flag to dataset versions API responses, I've removed them from the code here as well. This PR no longer depends on/is related to any dataverse PRs.

/

I've added tests and docs on how to use.

Yes

## Additional documentation:

/


docs/useCases.md
```diff
@@ -871,6 +872,36 @@ Note that `collectionId` is an optional parameter to filter datasets by collecti
...
+/* ... */
+```
+
+_See [use case](../src/datasets/domain/useCases/GetDatasetVersions.ts) implementation_.
+
+- The `datasetId` parameter can be a string, for persistent identifiers, or a number, for numeric identifiers.
+- **limit**: (number) Limit for pagination.
+- **offset**: (number) Offset for pagination.
+- **excludeMetadataBlocks**: (boolean) Exclude metadata blocks (default: false).
+- **excludesFiles**: (boolean) Exclude files (default: true).
```
ekraffmiller: I think returning the complete list of files in the response could be a problem - some of our datasets have several thousand files.  We have another use case, GetDatasetFiles, which returns the files with pagination.  My suggestion is to remove files from the response, and use the other use case to get the files with pagination.

ekraffmiller (changes requested):
Hi @vera, thanks for your PR. We don't currently need this use case in our SPA, because we use GetDatasetVersionsSummaries, and GetDatasetVersionDiff to display a list of Dataset versions.  But since it is available on the Dataverse API I think it's ok to add.  The only caveat I have is to not return the list of files, for performance reasons.

docs/useCases.md
vera: @ekraffmiller thanks for your review and sorry for the delay in getting back to this.

I have removed the `excludesFiles` param from this client implementation, so it's no longer possible to request the list of files without pagination.

I'm just wondering:

* Should I also add a note in the documentation of the use case that we are differing from the Dataverse API specification here, and why?
* Does this mean that the param should really be removed from the Dataverse API entirely? Or are there other use cases, outside of dataverse-client-javascript, which won't have this problem?

ekraffmiller: I think it's ok to differ from the Dataverse API, because it is similar to what we are doing in the GetDataset use case.  We have a separate use case, getDatasetFiles, that gets the files for a dataset version, with pagination.  It could be helpful, though to put a comment in the docs about that, so users know to use getDatasetFiles.  
We might want to remove this from the Dataverse API, but it will be a breaking change, so will have to think about that more.

src/datasets/domain/models/DatasetVersion.ts
```diff
@@ -0,0 +1,30 @@
...
+  versionState: DatasetVersionState
+  latestVersionPublishingState: DatasetVersionState
+  lastUpdateTime: string
+  releaseTime?: string
+  createTime: string
+  publicationDate: string
+  citationDate: string
+  license: DatasetLicense
+  fileAccessRequest: boolean
+  files?: Array<FilePayload>
```
ekraffmiller:
```suggestion
```
Since we are not returning files, this field can be removed from the model.

test/integration/datasets/DatasetsRepository.test.ts
```diff
@@ -1555,6 +1555,200 @@ describe('DatasetsRepository', () => {
...
+      expect(firstPage.versions[4].versionNumber).toBe(1)
+      expect(firstPage.versions[4].versionMinorNumber).toBe(7)
+
+      // Test that all versions are returned without pagination
+      const allVersions = await sut.getDatasetVersions(testDatasetIds.numericId)
+      expect(allVersions.versions.length).toBe(22) // 1 initial + 21 updates
+      expect(allVersions.totalCount).toBe(22)
+
+      await deletePublishedDatasetViaApi(testDatasetIds.persistentId)
+    }, 180000)
```
ekraffmiller:
```suggestion
      await deletePublishedDatasetViaApi(testDatasetIds.persistentId)
    }, 5000)
```
Can we shorten the time for this test? Fewer pages or fewer items per page?

docs/useCases.md
vera: OK, I've added a comment to the docs and linked the getDatasetFiles use case.

test/integration/datasets/DatasetsRepository.test.ts
vera: Sure. I had copied the number of versions to create and the timeout from another test ('should return dataset versions summaries with pagination'). I'm not sure if the 5000 timeout is sufficient for the test now, should I decrease any other numbers in the test?

I tried to check by running `npm run test:integration -- -t "getDatasetVersions"`, but it seems I'm unable to run the integration tests locally, because I get some unrelated error about a missing storage driver API:

```
            message: 'API endpoint does not exist on this server. Please check your code for typos, or consult our API guide at http://guides.dataverse.org.',
            requestUrl: 'http://localhost:8080/api/v1/admin/dataverse/testDatasetVersionsCollection/storageDriver',
            requestMethod: 'PUT'
          }
        },
        status: 404
      }

      at test/testHelpers/collections/collectionHelper.ts:144:13
          at Generator.throw (<anonymous>)

  ● DatasetsRepository › getDatasetVersions › should return dataset versions when dataset exists

    Error while setting storage driver for collection testDatasetVersionsCollection

      143 |   } catch (error) {
      144 |     console.log(error)
    > 145 |     throw new Error(`Error while setting storage driver for collection ${collectionAlias}`)
          |           ^
      146 |   }
      147 | }
      148 |

      at test/testHelpers/collections/collectionHelper.ts:145:11
          at Generator.throw (<anonymous>)
      at rejected (test/testHelpers/collections/collectionHelper.ts:6:65)
```

ekraffmiller (approved):
looks good, thanks for the updates

ekraffmiller:
The integration tests are failing because it doesn't contain the latest from the develop branch, would be good to merge that  when testing this.

ekraffmiller:
Hi @vera , can you update this PR with the latest from develop? I think that should fix the integration test.

vera:
> Hi @vera , can you update this PR with the latest from develop? I think that should fix the integration test.

@ekraffmiller I've updated the PR with develop, and also fixed some other mistakes which were causing the integration tests to fail (1. removed the `totalCount`, which I wrongly assumed the Dataverse API returns, but it doesn't actually do that, and 2. increased the timeout for the pagination test with 21 versions, because it had previously failed due to a timeout: https://github.com/IQSS/dataverse-client-javascript/actions/runs/32116134773/job/95645950539#step:5:456)

The tests are succeeding now.

I've also added a line to the changelog.

ChengShi-1:
Hi Vera, since currently the `GetVersion` api has a bug flagged in https://github.com/IQSS/dataverse.harvard.edu/issues/397 and we are not using it now for Modern UI, I'll put this On Hold untill we solve the bug

vera:
> Hi Vera, since currently the `GetVersion` api has a bug flagged in [IQSS/dataverse.harvard.edu#397](https://github.com/IQSS/dataverse.harvard.edu/issues/397) and we are not using it now for Modern UI, I'll put this On Hold untill we solve the bug

Hi, thanks for the feedback, since that bug seems to affect only Harvard Dataverse, would it be possible to review this anyways?

src/datasets/domain/models/DatasetVersion.ts
```diff
@@ -0,0 +1,28 @@
+import { DatasetLicense, DatasetMetadataBlocks, DatasetVersionState } from "./Dataset";
+import { FilePayload } from "../../../files/infra/repositories/transformers/FilePayload";
```
ChengShi-1, line 2: Could you remove `
import { FilePayload } from "../../../files/infra/repositories/transformers/FilePayload";` since its value is never read.

ChengShi-1 (commented):
Hi @vera , I just left some comments below. Thanks for checking them

src/datasets/domain/models/DatasetVersion.ts
```diff
@@ -0,0 +1,28 @@
...
+  versionMinorNumber?: number
+  internalVersionNumber: number
+  versionState: DatasetVersionState
+  latestVersionPublishingState: DatasetVersionState
+  lastUpdateTime: string
+  releaseTime?: string
+  createTime: string
+  publicationDate: string
+  citationDate: string
+  license: DatasetLicense
```
ChengShi-1, line 21: license, publicationDate should be optional

Could we reuse the existing `Dataset` model(or other existing model) and version transformer here? `Dataset` already represents a specific dataset version. This new interface duplicates much of it while exposing raw API field names and introducing inconsistent types. If `excludeMetadataBlocks` prevents reuse, could we derive a type from `Dataset` instead?

src/datasets/infra/repositories/DatasetsRepository.ts
```diff
@@ -372,6 +373,39 @@ export class DatasetsRepository extends ApiRepository implements IDatasetsReposi
...
+      queryParams.set('excludeMetadataBlocks', excludeMetadataBlocks.toString())
+    }
+
+    return this.doGet(
+      this.buildApiEndpoint(this.datasetsResourceName, 'versions', datasetId),
+      true,
+      queryParams
+    )
+      .then((response) => ({
+        versions: response.data.data
```
ChengShi-1, line 402: Returned metadata does not match its declared type. Raw `metadataBlocks` is an object, 
```
metadataBlocks: {
  citation: {
    fields: [
      { typeName: "title", value: "My dataset" }
    ]
  }
}
```

but `DatasetMetadataBlocks`  expects an array
```
metadataBlocks: [
  {
    name: "citation",
    fields: {
      title: "My dataset"
    }
  }
]
```

The new method returns Dataverse’s response directly, without converting it TypeScript therefore allows:`version.metadataBlocks?.map(block => block.name)`

vera: I've updated the code so the Dataverse API response is sent through a transformer to return the correct type.

src/datasets/domain/models/DatasetVersion.ts
vera: Yes, I've updated the code to use a type derived from `Dataset`.

src/datasets/domain/models/DatasetVersion.ts
vera: Done :)

docs/useCases.md
```diff
@@ -1262,6 +1263,36 @@ Note that `collectionId` is an optional parameter to filter datasets by collecti
 
 The `DatasetPreviewSubset`returned instance contains a property called `totalDatasetCount` which is necessary for pagination.
 
+#### Get Dataset Versions
+
+Returns the total count of versions and an array of [Dataset](../src/datasets/domain/models/Dataset.ts) that contains information about every specific version.
```
ChengShi-1: Could you change to `list of versions`  instead of `total count`? It will not return any count so a list of versions is more accurate

ChengShi-1 (commented):
Thanks for the fix! Just one small doc change. Otherwise, looks good to me!

docs/useCases.md
vera: Done!

ChengShi-1 (approved):
Thanks for the fix! Looks good to me

PR #418, opened 2026-01-27
Add use case for getting list of collections a user may create datasets in

vera:

This PR adds code for [getting the list of collections a user may create datasets in](https://guides.dataverse.org/en/latest/api/native-api.html#mydata-collection-list). (https://github.com/IQSS/dataverse/pull/11681)

/

## Related Dataverse PRs:

/

/

I've added tests and docs on how to use.

Yes

## Additional documentation:

/


ekraffmiller:
Hi @vera thanks for your PR!  I was just reviewing the API that this use case is wrapping, and it has some issues with it -  it doesn't have a paging option, and if the endpoint is called for a user such as an admin that returns a long list of collections, the performance is not good.  
We have a more general API, /mydata/retrieve, that takes parameters for  paging, dvObject type, roles, and also allows filtering of the results by collection name.  It's the API used in the MyData page to filter the table contents. With the right parameters, it can return a list of collections that a user can create a dataset in. Could that be used for your purposes instead?  We are thinking we might deprecate the collectionList endpoint in favor of mydata/retrieve.

UPDATE: after talking to the team, we think it's better to keep collectionList as a separate API, and add pagination and filtering to it.  I created an issue for it here: https://github.com/IQSS/dataverse/issues/12423

ekraffmiller:
Putting this on hold, so it can be updated when https://github.com/IQSS/dataverse/issues/12423 is completed

vera:
@pdurbin has let me know that a PR for https://github.com/IQSS/dataverse/issues/12423 (https://github.com/IQSS/dataverse/pull/12436) is in the works. I'll update this PR with the new pagination and filtering params once that's merged. For now, I've just fixed merge conflicts.

ChengShi-1:
Hi @vera , I thought this was Ready for Review, so I assigned Copilot to review it, sorry for interrupting! I see your comment here and realize you're still working on this PR now.

> @pdurbin has let me know that a PR for [IQSS/dataverse#12423](https://github.com/IQSS/dataverse/issues/12423) ([IQSS/dataverse#12436](https://github.com/IQSS/dataverse/pull/12436)) is in the works. I'll update this PR with the new pagination and filtering params once that's merged. For now, I've just fixed merge conflicts.

PR #433, opened 2026-03-20
fix client package self-reference and alpha publish workflow

ekraffmiller:

- Removes the self-dependency from @IQSS/dataverse-client-javascript
- Fxes GitHub Packages publish workflows to use the correct lowercase package name and the current base version for alpha releases.

- Closes #

## Related Dataverse PRs:

- Depends on #


ChengShi-1 (approved):
looks good

PR #438, opened 2026-03-31
AccessRepository.ts omits credentials if using Bearer Token Authorization

### LINKED ISSUE #437: Don't send session cookies when using Bearer Token authentication
In testing this frontend PR https://github.com/IQSS/dataverse-frontend/issues/896 we are getting different behavior depending on the existence of dataverse session cookies.  For example, when an API call is made from the SPA and the user is not logged in, the Download with Guestbook API call is using session cookies to authenticate the user.  

To prevent this, we need to modify the API calls so that if bearer token authentication is configured, no session cookies are sent in the API call.

ekraffmiller:
The Access API for download sometimes uses session cookies for authentication, which can be a problem if the user is running JSF and SPA in the same browser.  To work around this, this change omits session cookies for all API calls in AccessRepository.

- Closes #437

## Related Dataverse PRs:

(https://github.com/IQSS/dataverse/issues/12267)

Review tests


ChengShi-1 (approved):
Looks good

PR #440, opened 2026-04-20
Setup TS change to allow skipping the containers setup.

jp-tosca:

When running tests, containers are started to run against but there are use cases like development or more complicated configurations that would be helpful to run against local installations that have been set up in a different way.

You can run a test for example with 
```npm run test GetCollection``` 
or 
```SKIP_CONTAINERS=true npm run test GetCollection```

You could also set ```const SKIP_CONTAINERS``` in setup.ts


rtreacy (approved):
Looks good

ekraffmiller:
looks good, merging 👍

PR #443, opened 2026-04-23
tests fix

jp-tosca:

There was a change on DV with https://github.com/IQSS/dataverse/pull/12145 that caused tests to fail, this is meant to be a temporary patch since it seems there is a bug with this PR.

We did a few testing and provided a .tsv file and received .tab response.
When providing a .tab file we received a .tsv response.
=


ChengShi-1:
Thanks for the fix, I have manually added to new dev branch [here](https://github.com/IQSS/dataverse-client-javascript/blob/1dc5d8b32dcc4e1e017bc8d10a4acf142ef4e400/test/integration/collections/CollectionsRepository.test.ts#L1000)

PR #449, opened 2026-05-11
Extend get guestbooks use case with stats and Download responses

### LINKED ISSUE #448: Add usage and response counts to get guestbooks use case
Related PR: https://github.com/IQSS/dataverse/pull/12269

Currently, the GET Guestbook API returns metadata about guestbooks (e.g., name, required fields, custom questions), but it does not include usage and responses, needed for Manage Guestbooks table

ChengShi-1:
- Get Guestbooks By Collection Id: Added optional `includeStats` and `includeInherited` support to `getGuestbooksByCollectionId`.
- write use cases of downloading all responses of a collection or download responses of a guestbook

- Closes #448 

## Related Dataverse PRs:

- Depends on #https://github.com/IQSS/dataverse/pull/12326 and https://github.com/IQSS/dataverse-client-javascript/issues/448

Test failed about storage drive, so I also copy pastes a `CollectionHelper.ts` from this PR. waiting to be merged
https://github.com/IQSS/dataverse-client-javascript/pull/431/changes#diff-edac108c1a43d5773116d19797d04ecdf890fc97b5cf83ce2d56d38b9a7aca18


src/guestbooks/domain/useCases/DownloadGuestbookResponsesByDataverseId.ts
```diff
@@ -0,0 +1,20 @@
+import { UseCase } from '../../../core/domain/useCases/UseCase'
+import { IGuestbooksRepository } from '../repositories/IGuestbooksRepository'
+
+export class DownloadGuestbookResponsesByDataverseId implements UseCase<string> {
```
ekraffmiller: We should use "collection" in the use case and all the comments, rather than dataverse

src/guestbooks/domain/useCases/GetGuestbookResponsesOfAGuestbook.ts
```diff
@@ -0,0 +1,21 @@
...
+import { GuestbookResponse } from '../models/GuestbookResponse'
+import { IGuestbooksRepository } from '../repositories/IGuestbooksRepository'
+
+export class GetGuestbookResponsesOfAGuestbook implements UseCase<GuestbookResponse[]> {
+  constructor(private readonly guestbooksRepository: IGuestbooksRepository) {}
+
+  /**
+   * Returns guestbook responses for one guestbook in a dataverse collection.
+   *
+   * @param {number | string} dataverseId - Dataverse identifier.
```
ekraffmiller: The API for getting the responses for a guestbook in JSON form is, for example: 
curl -H  "X-Dataverse-key:$API_TOKEN" "$SERVER_URL/api/guestbooks/$ID/responses?limit10&offset=0"
So we need to add parameters for limit and offset.
Another thing I noticed though, is that this API doesn't allow us to pass a dataverseId, so it's going to return all guestbook responses regardless of collection. So unfortunately I think we to ask for another change to the API -  to be able to pass guestbookId and collectionId and get all the responses for a guestbook within a specific collection, returned in JSON format.

src/guestbooks/domain/useCases/DownloadGuestbookResponsesOfAGuestbook.ts
```diff
@@ -0,0 +1,24 @@
+import { UseCase } from '../../../core/domain/useCases/UseCase'
+import { IGuestbooksRepository } from '../repositories/IGuestbooksRepository'
+
+export class DownloadGuestbookResponsesOfAGuestbook implements UseCase<string> {
+  constructor(private readonly guestbooksRepository: IGuestbooksRepository) {}
+
+  /**
```
ekraffmiller, line 7: this use case should use "collection" instead of dataverse, in the documentation and declaration

src/guestbooks/domain/useCases/GetGuestbookResponsesByDataverseId.ts
```diff
@@ -0,0 +1,17 @@
+import { UseCase } from '../../../core/domain/useCases/UseCase'
+import { GuestbookResponse } from '../models/GuestbookResponse'
+import { IGuestbooksRepository } from '../repositories/IGuestbooksRepository'
+
+export class GetGuestbookResponsesByDataverseId implements UseCase<GuestbookResponse[]> {
```
ekraffmiller: I don't think this use case is needed, we only need to download the csv version of the responses, we don't need an array of JSON for displaying.

ekraffmiller (commented):
Hi @ChengShi-1 I left some comments on the PR, hope it's not too confusing! After going back on forth between the use cases and the API, I think there is actually one more change we need in the API - which I describe in my comment in GetGuestbookResponsesOfAGuestbook.  I'm happy to discuss more about it.

src/guestbooks/domain/useCases/GetGuestbookResponsesOfAGuestbook.ts
ChengShi-1: I created an issue in dataverse for the api missing https://github.com/IQSS/dataverse/issues/12451

docs/useCases.md
```diff
@@ -3168,21 +3171,88 @@ _See [use case](../src/guestbooks/domain/useCases/GetGuestbook.ts) implementatio
...
+##### Example call:
+
+```typescript
+import { getGuestbookResponsesByGuestbookId } from '@iqss/dataverse-client-javascript'
+
+const guestbookId = 123
+const limit = 10
+const offset = 0
+
+getGuestbookResponsesByGuestbookId
```
ekraffmiller, line 3206: The API will return pagination info as well, so the return type should be GustbookSubset, containing the array of Guestbook items, and the total count, similar to how we handle pagination for other objects

ekraffmiller (commented):
Looks good! just need to update the return type of GetGuestbookResponsesByGuestbookId to support pagination.

docs/useCases.md
ChengShi-1: Added. Thanks for catching this😃

ekraffmiller (approved):
thanks for the updates, looks good!

PR #457, opened 2026-06-16
Get Dataset Reviews use case

### LINKED ISSUE #447: New Use Case: get local reviews of Dataset
For the Dataset page, we want to add a section that lists the local reviews of a dataset, if any are available.  
For this use case, the parameter to send the API is the persistentId of the dataset, and it will return a list of reviews for that dataset.

ekraffmiller:
Adds a use case for getting dataset reviews

- Closes #447 

The integration test includes setup code that adds the review.tsv metadatablock, and updates the solar schema. This is required because the API endpoint uses solr search to find the reviews related to a dataset.

yes, an item as added to the CHANGELOG.md


ChengShi-1 (approved):
looks good

